PR #8, opened 2020-09-08
Add export routes

nguyen102:
Issue #, if available:

Description of changes:
* Added `BadRequestError` so we can inform the user why the request is invalid.
* Added handling of `TooManyRequestsError`
* Added handling export routes (system, patient, group)

I want to check if we're ok with the approach in this PR to create a new route and handler for export. The alternative is to add these routes to `rootRoute` and `genericResourceRoute`. However, I feel this alternative would be more confusing as `export` would be in two different places instead of one. 

Assuming we're ok with the current PR, I'll continue the next revision of this PR with unit tests.

Related PR:
https://github.com/awslabs/fhir-works-on-aws-interface/pull/13
https://github.com/awslabs/fhir-works-on-aws-authz-rbac/pull/7

- [x] Added tests

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -63,8 +63,8 @@ export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGeneri
     app.use('/metadata', metadataRoute.router);
 
     // Export
-    if (fhirConfig.profile.genericResource?.persistence) {
-        const exportRoute = new ExportRoute(serverUrl, fhirConfig.profile.genericResource?.persistence);
+    if (fhirConfig.profile.export) {
+        const exportRoute = new ExportRoute(serverUrl, fhirConfig.profile.export);
         app.use('/', exportRoute.router);
     }
 
@@ -127,9 +127,7 @@ export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGeneri
     }
 
     app.use(applicationErrorMapper);
-    app.use((err: any, req: express.Request, res: express.Response, next: express.NextFunction) => {
-        return httpErrorHandler(err, req, res, next, fhirConfig.defaultRetryRequestInSeconds);
-    });
+    app.use(httpErrorHandler);
     app.use(unknownErrorHandler);
 
     return app;
--- a/src/app.ts
+++ b/src/app.ts
@@ -42,16 +42,13 @@ export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGeneri
     app.use(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
         try {
             const requestInformation = getRequestInformation(req.method, req.path);
-            console.log('requestInformation', requestInformation);
             const accessToken: string = cleanAuthHeader(req.headers.authorization);
             const isAllowed: boolean = fhirConfig.auth.authorization.isAuthorized({
                 ...requestInformation,
                 accessToken,
             });
-            // const isAllowed = true;
             if (isAllowed) {
-                const requesterUserId = fhirConfig.auth.authorization.getRequesterUserId(accessToken);
-                res.locals.requesterUserId = requesterUserId;
+                res.locals.requesterUserId = fhirConfig.auth.authorization.getRequesterUserId(accessToken);
                 next();
             } else {
                 res.status(403).json({ message: 'Forbidden' });
@@ -66,6 +63,7 @@ export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGeneri
     app.use('/metadata', metadataRoute.router);
 
     // Export
+
     if (fhirConfig.profile.genericResource?.persistence) {
         const exportRoute = new ExportRoute(serverUrl, fhirConfig.profile.genericResource?.persistence);
         app.use('/', exportRoute.router);
@@ -130,7 +128,10 @@ export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGeneri
     }
 
     app.use(applicationErrorMapper);
-    app.use(httpErrorHandler);
+    app.use((err: any, req: express.Request, res: express.Response, next: express.NextFunction) => {
+        return httpErrorHandler(err, req, res, next, fhirConfig.defaultRetryRequestInSeconds);
+    });
+    // app.use(httpErrorHandler);
     app.use(unknownErrorHandler);
 
     return app;
--- a/src/app.ts
+++ b/src/app.ts
@@ -130,7 +130,6 @@ export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGeneri
     app.use((err: any, req: express.Request, res: express.Response, next: express.NextFunction) => {
         return httpErrorHandler(err, req, res, next, fhirConfig.defaultRetryRequestInSeconds);
     });
-    // app.use(httpErrorHandler);
     app.use(unknownErrorHandler);
 
     return app;
--- a/src/app.ts
+++ b/src/app.ts
@@ -63,8 +63,8 @@ export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGeneri
     app.use('/metadata', metadataRoute.router);
 
     // Export
-    if (fhirConfig.profile.export) {
-        const exportRoute = new ExportRoute(serverUrl, fhirConfig.profile.export);
+    if (fhirConfig.profile.bulkDataAccess) {
+        const exportRoute = new ExportRoute(serverUrl, fhirConfig.profile.bulkDataAccess);
         app.use('/', exportRoute.router);
     }
 
--- a/src/app.ts
+++ b/src/app.ts
@@ -67,7 +67,7 @@ export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGeneri
 
     // Export
     if (fhirConfig.profile.genericResource?.persistence) {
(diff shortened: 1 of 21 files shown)


src/router/routes/exportRoute.ts
```diff
@@ -0,0 +1,106 @@
...
+        const requestQueryParams: any = {};
+        // eslint-disable-next-line no-unused-expressions
+        isString(req.query._outputFormat) ? (requestQueryParams._outputFormat = req.query._outputFormat) : '';
+        // eslint-disable-next-line no-unused-expressions
+        Number(req.query._since) ? (requestQueryParams._since = Number(req.query._since)) : '';
+        // eslint-disable-next-line no-unused-expressions
+        isString(req.query._type) ? (requestQueryParams._type = req.query._type) : '';
+
+        if (requestQueryParams._outputFormat && requestQueryParams._outputFormat !== 'ndjson') {
+            throw new BadRequestError('We only support exporting resources into ndjson formatted file');
```
carvantes: Use `http-errors` instead
```js
import createError from 'http-errors';
throw new createError.BadRequest(message);
```

The routing layer can throw http errors. There's no need to create a custom error like `BadRequestError`

carvantes (changes requested):

src/router/routes/exportRoute.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    async initiateExportRequests(req: express.Request, res: express.Response, requestGranularity: ExportType) {
+        const requestQueryParams: any = {};
+        // eslint-disable-next-line no-unused-expressions
+        isString(req.query._outputFormat) ? (requestQueryParams._outputFormat = req.query._outputFormat) : '';
+        // eslint-disable-next-line no-unused-expressions
+        Number(req.query._since) ? (requestQueryParams._since = Number(req.query._since)) : '';
+        // eslint-disable-next-line no-unused-expressions
+        isString(req.query._type) ? (requestQueryParams._type = req.query._type) : '';
```
carvantes: It's weird to do assignments like this on a ternary. The usual is:
```js
requestQueryParams._outputFormat = isString(req.query._outputFormat) ? req.query._outputFormat : "";
```

I don't think that there's a good reason to use `// eslint-disable-next-line` here

carvantes (changes requested):

src/router/routes/exportRoute.ts
nguyen102: Good idea. fixed

src/router/routes/exportRoute.ts
nguyen102: fixed

src/app.ts
```diff
@@ -60,6 +62,12 @@ export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGeneri
     const metadataRoute: MetadataRoute = new MetadataRoute(fhirVersion, configHandler);
     app.use('/metadata', metadataRoute.router);
 
+    // Export
```
rsmayda, line 65: support for export should be optional similar to Bundles

src/router/routes/exportRoute.ts
```diff
@@ -0,0 +1,81 @@
...
+            .status(202)
+            .send();
+    }
+
+    init() {
+        // Start export job
+        this.router.get(
+            '/\\$export',
+            RouteHelper.wrapAsync(async (req: express.Request, res: express.Response) => {
+                const requestGranularity: ExportType = 'system';
```
rsmayda: change to be 'exportType'

src/router/routes/exportRouteHelper.ts
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable no-underscore-dangle */
+import express from 'express';
+import { ExportType, InitiateExportRequest } from 'fhir-works-on-aws-interface';
+import createHttpError from 'http-errors';
+import isString from 'lodash/isString';
+
+export default class ExportRouteHelper {
+    static buildInitiateExportRequest(req: express.Request, res: express.Response, requestGranularity: ExportType) {
```
rsmayda:
```suggestion
    static buildInitiateExportRequest(req: express.Request, res: express.Response, exportType: ExportType) {
```

src/app.ts
nguyen102: Good idea. Adding that functionality.

src/router/routes/exportRouteHelper.ts
nguyen102: fixed

src/router/routes/exportRoute.ts
nguyen102: fixed

src/regExpressions.ts
```diff
@@ -6,7 +6,7 @@
 export const uuidRegExp = /\w{8}-\w{4}-\w{4}-\w{4}-\w{12}/;
 export const resourceTypeWithUuidRegExp = /\w+\/\w{8}-\w{4}-\w{4}-\w{4}-\w{12}/;
 
-export const utcTimeRegExp = /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d+Z/;
+export const utcTimeRegExp = /\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d+)?Z/;
```
rsmayda, line 9: Whats the diff here?

src/router/routes/errorHandling.ts
```diff
@@ -22,10 +23,22 @@ export const applicationErrorMapper = (
         next(new createError.BadRequest(`Failed to parse request body as JSON resource. Error was: ${err.message}`));
         return;
     }
+    if (err instanceof TooManyConcurrentExportRequestsError) {
+        next(
+            new createError.TooManyRequests(
+                'There is currently too many concurrent export requests. Please try again later',
```
rsmayda: Lets make this generic
```suggestion
                'There is currently too many requests. Please try again later',
```

src/router/routes/errorHandling.ts
```diff
@@ -22,10 +23,22 @@ export const applicationErrorMapper = (
...
+            ),
+        );
+        return;
+    }
     next(err);
 };
 
 export const httpErrorHandler = (err: any, req: express.Request, res: express.Response, next: express.NextFunction) => {
+    if (err instanceof createError.TooManyRequests) {
+        const RETRY_AGAIN_IN_SECONDS = 15 * 60;
```
rsmayda: Is 15 mins the suggested or how did we come up with that? Add a minor comment saying 15mins. 
```suggestion
        const RETRY_AGAIN_IN_SECONDS = 15 * 60; // 15 minutes
```

rsmayda (commented):
Do we need to update the capability statement at all?

src/regExpressions.ts
nguyen102: It now allows for the time stamp to not include seconds. So
`2020-09-01T12:00:00.123Z` and `2020-09-01T12:00:00Z` are both valid.

src/router/routes/errorHandling.ts
nguyen102: Updated

src/router/routes/errorHandling.ts
nguyen102: Updated with the comment. I got 15 minutes from reading forums online where people did ETL jobs with Glue. It usually took them under 10 minutes, so I thought 15 minutes would be a safe number for the retry.

src/regExpressions.ts
carvantes: A FHIR instant is not necessarily UTC. It SHALL include a time zone
https://www.hl7.org/fhir/datatypes.html

src/router/routes/exportRoute.ts
```diff
@@ -0,0 +1,80 @@
...
+
+/* eslint-disable no-underscore-dangle */
+import express, { Router } from 'express';
+import { BulkDataAccess, ExportType, InitiateExportRequest } from 'fhir-works-on-aws-interface';
+import createHttpError from 'http-errors';
+import RouteHelper from './routeHelper';
+import ExportHandler from '../handlers/exportHandler';
+import ExportRouteHelper from './exportRouteHelper';
+
+export default class ExportRoute {
```
carvantes, line 14: I don't think this should be a class. 

It could just be a function like:
```
(serverUrl: string, bulkDataAccess: BulkDataAccess) => expressRouter
```
This class only has a public property that we care about (`router`) and a big constructor. It's not a good pattern.

I know that all other Routes follow this pattern but I don't think we should continue to do that.

nguyen102:
https://github.com/awslabs/fhir-works-on-aws-routing/pull/8#pullrequestreview-486340824

Added capability statement

src/regExpressions.ts
nguyen102: Adding code to allow for timezones.

src/router/routes/exportRoute.ts
nguyen102: That makes sense. I think all the router class should follow the same pattern, whether it should be a function or a class. Therefore, I think if we change this class to a function, we should do with the other classes also. Can we leave this task as a future refactoring task?

carvantes: yes, it can be done on a separate PR

carvantes (approved):

PR #9, opened 2020-09-14
Handle get-export-status responses

nguyen102:
Description of changes:
* Added code to handle the different responses for an export status request.

**Related PRs**
* https://github.com/awslabs/fhir-works-on-aws-interface/pull/14
* https://github.com/awslabs/fhir-works-on-aws-authz-rbac/pull/8

### DIFF
--- a/src/router/handlers/exportHandler.ts
+++ b/src/router/handlers/exportHandler.ts
@@ -4,6 +4,7 @@
  */
 
 import {
+    AccessBulkDataJobRequest,
     Authorization,
     BulkDataAccess,
     GetExportStatusResponse,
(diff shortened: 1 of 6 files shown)


src/router/routes/exportRoute.ts
```diff
@@ -59,17 +59,43 @@ export default class ExportRoute {
 
         // Export Job Status
         this.router.get(
-            '/\\$export/:jobId',
+            '/\\$export/:requesterUserId/:jobId',
```
rsmayda: Hmm is this expected part of the spec?

nguyen102: The FHIR spec didn't specify what the job status URL be, so this is up for us to decide. I'm including this so we can use it to authorized the user. We get the user's `userId` from their `accessToken` and compare it to the `requesterUserId` here to confirm that they are the same user. Given that they are authorized to get this job's status.

rsmayda: Let's talk offline on a good solution

src/router/handlers/exportHandler.ts
```diff
@@ -16,11 +16,11 @@ export default class ExportHandler {
         return this.bulkDataAccess.initiateExport(initiateExportRequest);
```
rsmayda, line 26: So looking at the code for bundles it looks like AuthZ still happens in the routing layer: https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/router/bundle/bundleHandler.ts#L127

Thoughts on mimicing this approach?

rsmayda (approved):

src/router/handlers/exportHandler.ts
```diff
@@ -34,13 +35,19 @@ export default class ExportHandler {
     async cancelExport(jobId: string, requesterUserId: string): Promise<void> {
         const jobDetails = await this.bulkDataAccess.getExportStatus(jobId);
         await this.checkIfRequesterHasAccessToJob(jobDetails, requesterUserId);
+        if (['completed', 'failed'].includes(jobDetails.jobStatus)) {
```
rsmayda, line 38: I think the only time you can cancel the request is when it is in 'pending' right?

nguyen102: The different type of `jobStatus` are

` 'completed' | 'failed' | 'in-progress' | 'canceled' | 'canceling'`

If the job is currently in `'in-progress' | 'canceled' | 'canceling'`, we don't want to throw an error for the cancel request. We can just proceed as is.

rsmayda: What will the service do when the job in canceled gets asked to cancel again? Basically could we return from here?

PR #11, opened 2020-09-15
Bugfix: add missing package

### LINKED ISSUE #10: Missing Express types
Hi,
I'm hitting a transpile error when using the routing npm package without the other fhir-works components:
```
node_modules/fhir-works-on-aws-routing/lib/app.d.ts:2:119 - error TS2307: Cannot find module 'express-serve-static-core' or its corresponding type declarations.

2 export declare function generateServerlessRouter(fhirConfig: FhirConfig, supportedGenericResources: string[]): import("express-serve-static-core").Express; 
```
Just need to add `@types/express-serve-static-core` into the dependencies.
Thanks!
Allan

SanketD92:
Issue #, if available:
https://github.com/awslabs/fhir-works-on-aws-routing/issues/10

Description of changes:
Running "yarn install" failed for this repository and required this package addition to the dependencies to fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
     "lodash": "^4.17.15",
     "mime-types": "^2.1.26",
     "serverless-http": "^2.3.1",
-    "uuid": "^3.4.0"
+    "uuid": "^3.4.0",
+    "@types/express-serve-static-core": "^4.17.2"
   },
   "devDependencies": {
     "@types/errorhandler": "^0.0.32",
--- a/src/app.ts
+++ b/src/app.ts
@@ -20,7 +20,7 @@ import { applicationErrorMapper, httpErrorHandler, unknownErrorHandler } from '.
 
 const configVersionSupported: ConfigVersion = 1;
 
-export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGenericResources: string[]) {
+export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGenericResources: string[]): any {
     if (configVersionSupported !== fhirConfig.configVersion) {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -37,7 +37,8 @@
     "lodash": "^4.17.15",
     "mime-types": "^2.1.26",
     "serverless-http": "^2.3.1",
-    "uuid": "^3.4.0"
+    "uuid": "^3.4.0",
+    "@types/express-serve-static-core": "^4.17.2"
```
rsmayda, line 41: This is a dev dependency

SanketD92: I thought so too, but when I added it as a devDependency it still failed on yarn install. When I moved it to dependencies, it didn't fail anymore. Am I missing a step?

rsmayda: Whoops your right: https://stackoverflow.com/a/46011417/9041033

rsmayda: I would suggest explicitly calling out the return type here: https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/app.ts#L23

src/app.ts
```diff
@@ -20,7 +20,7 @@ import { applicationErrorMapper, httpErrorHandler, unknownErrorHandler } from '.
 
 const configVersionSupported: ConfigVersion = 1;
 
-export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGenericResources: string[]) {
+export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGenericResources: string[]): any {
```
nguyen102: Can you check if the type here is supposed to be `Express`?

And then at the top
`import express, { Express } from 'express';`

nguyen102 (approved):

PR #13, opened 2020-09-24
feat: Pass down allowed resource types to search service

carvantes:
Description of changes:

Call `auth.resourceTypesAllowedForOperation()` and pass down the results to search as described on https://github.com/awslabs/fhir-works-on-aws-interface/pull/16

Note: Need to publish the updated interface to npm first. Until then, checks are failing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "ajv": "^6.11.0",
     "errorhandler": "^1.5.1",
     "express": "^4.17.1",
-    "fhir-works-on-aws-interface": "^1.0.0",
+    "fhir-works-on-aws-interface": "^2.0.0",
     "flat": "^5.0.0",
     "http-errors": "^1.8.0",
     "lodash": "^4.17.15",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fhir-works-on-aws-routing",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "description": "FHIR Works on AWS routing implementation",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
@@ -28,10 +28,10 @@
     "prepublish": "tsc"
   },
   "dependencies": {
-    "fhir-works-on-aws-interface": "^1.0.0",
     "ajv": "^6.11.0",
     "errorhandler": "^1.5.1",
     "express": "^4.17.1",
+    "fhir-works-on-aws-interface": "^1.0.0",
     "flat": "^5.0.0",
     "http-errors": "^1.8.0",
     "lodash": "^4.17.15",
--- a/src/app.ts
+++ b/src/app.ts
@@ -42,15 +42,11 @@ export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGeneri
         try {
             const requestInformation = getRequestInformation(req.method, req.path);
             const accessToken: string = cleanAuthHeader(req.headers.authorization);
-            const isAllowed: boolean = fhirConfig.auth.authorization.isAuthorized({
+            const isAllowed: boolean = await fhirConfig.auth.authorization.isAuthorized({
                 ...requestInformation,
                 accessToken,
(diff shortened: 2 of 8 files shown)


src/app.ts
```diff
@@ -47,6 +47,10 @@ export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGeneri
                 accessToken,
             });
             if (isAllowed) {
+                res.locals.allowedResourceTypes = fhirConfig.auth.authorization.getAllowedResourceTypesForOperation({
+                    operation: requestInformation.operation,
+                    accessToken,
+                });
```
rsmayda: Should we only call this if we need this (include/revinclude Search)? Perhaps we could use outgoing chaining? Like how we are doing the error handling (not sure if possible, to follow that express paradigm here)

Something like this middle ware: https://dev.to/getd/express-middleware-is-just-a-fancy-way-of-chaining-a-bunch-of-functions-explained-in-3-mins-43jf

src/router/handlers/resourceHandler.test.ts
```diff
@@ -365,7 +365,7 @@ describe('Testing search', () => {
         });
 
         // OPERATE
-        const searchResponse: any = await resourceHandler.typeSearch('Patient', { name: 'Henry' });
+        const searchResponse: any = await resourceHandler.typeSearch('Patient', { name: 'Henry' }, []);
```
rsmayda: We should probably have tests with the allowedResourceType?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fhir-works-on-aws-routing",
-  "version": "1.0.0",
+  "version": "1.1.0",
```
rsmayda, line 3: nit add changelog Change

rsmayda (commented):
Overall looks good only major question is if we can do the resource filter as a form of middleware?

package.json
carvantes: added

src/router/handlers/resourceHandler.test.ts
carvantes: yeah, this method is mostly a proxy to `searchService.typeSearch` which is mocked so the value of allowedResourceType doesn't really matter. 
I just added an assertion to verify that `allowedResourceType` is forwarded to `searchService.typeSearch`

src/app.ts
carvantes: I refactored it so that is called exclusively by search requests.

I don't think that resource filtering as middleware is a great idea. I don't like the idea of having the router being concerned with filtering search results. Plus, it is much more efficient to send the allowedResourceTypes to search since that can optimize the query plan(i.e. wildcards no longer need to query ALL resources. only the ones that are allowed)

src/router/routes/genericResourceRoute.ts
```diff
@@ -96,10 +99,16 @@ export default class GenericResourceRoute {
                     // Get the ResourceType looks like '/Patient'
                     const resourceType = req.baseUrl.substr(1);
                     const searchParamQuery = req.query;
+
+                    const allowedResourceTypes = await this.authService.getAllowedResourceTypesForOperation({
+                        operation: 'search-type',
+                        accessToken: cleanAuthHeader(req.headers.authorization),
+                    });
```
rsmayda, line 106: Do we need to add this to any bundle logic ie if you post a bundle that does a search?

carvantes: Not right now.

batch/transaction can include search queries, so passing down the allowedResourceTypes will be necessary.

However, we currently do not support batch: https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/router/bundle/bundleHandler.ts#L59

And our transaction implementation does not support search: https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/router/bundle/bundleParser.ts#L77

rsmayda (approved):

PR #14, opened 2020-09-30
feat: Support for CORS headers

AllanHodkinson:
Issue #, if available:
https://github.com/awslabs/fhir-works-on-aws-routing/issues/15

Description of changes:
Adding the Express CORS middleware to the router for compatibility with web applications.  Server metadata is updated to reflect this.  CORS middleware is using default configuration.  


rsmayda (commented):
Hey! we are trying to stay away from adding new env variables to non-deployment packages. Here is a PR (which is based off of yours) trying to achieve the same thing. Let us know your thoughts: https://github.com/awslabs/fhir-works-on-aws-routing/pull/18

I think taking some of the approaches we brought up with what you currently have will make this PR work!

src/app.ts
```diff
@@ -36,7 +37,10 @@ export function generateServerlessRouter(fhirConfig: FhirConfig, supportedGeneri
             limit: '6mb',
         }),
     );
-
+    // Add cors handler before auth to allow pre-flight requests without auth.
+    // TODO update fhirConfig.server to supply cors options? Env variable offered as an alternative
+    const corsOptions: CorsOptions = process.env.CORS_OPTIONS ? JSON.parse(process.env.CORS_OPTIONS) : {};
```
rsmayda: Could this be passed in as an optional parameter instead of a environment variable?

src/router/metadata/cap.rest.security.template.ts
```diff
@@ -8,7 +8,7 @@ import { Auth } from 'fhir-works-on-aws-interface';
 export default function makeSecurity(authConfig: Auth) {
     if (authConfig.strategy.service) {
         let security = {
-            cors: false,
+            cors: true,
```
rsmayda: This should be a variable passed into this function; see: https://github.com/awslabs/fhir-works-on-aws-routing/pull/18/files

rsmayda (commented):
Have you tested the 'in-flight' CORS use-case? It looks like it may require a change to the serverless.yaml

https://www.serverless.com/blog/cors-api-gateway-survival-guide#cors-preflight-requests.

src/router/metadata/cap.rest.security.template.ts
AllanHodkinson: Hi Robert,
I think this needs to be a constant - once we have the Cors middleware package in the router, cors headers will always be available in requests served the router.  With no config, the middleware will provide a basic setup; adding a CorsOptions will allow users to customize the settings if required.
A real-world FHIR server will need to provide cors headers so I don't think there is much value in making this configurable.  If it was configurable we'd also need to dynamically switch in the middleware according to this setting.

src/app.ts
AllanHodkinson: Hi Robert,
Yes I agree, I made this change in our code base so I could pass in a config at deployment time.
The correct way to do it is for the interface module to provide a CorOptions structure in fhirConfig.server.  I was under the impression that you had done this, but if not I can take it on?  We need this change in place (with a new npm package) before we can make the router changes - what is the procedure for doing this?

src/router/metadata/cap.rest.security.template.ts
rsmayda: Yeah totally understand your point here. One thing to note is that the capability statement is generated dynamically for every `/metadata` invocation based on the `config`.  Here is the metadata router: https://github.com/awslabs/fhir-works-on-aws-routing/blob/d120da1c0c3502c19f8da014335c95c840a787e2/src/router/routes/metadataRoute.ts#L29

So having the `isCorsEnabled` variable be passed in aligns with passing in of the `corsOptions` object

We want to be able to have CORS and no CORS - unless somewhere in the FHIR spec specifies CORS is required

src/app.ts
rsmayda: Yeah - I started to then decided not to. The `CorsOptions` object type comes from the **express** CORS middleware. We don't want any implementation detail leaking into the `interface`. This is to ensure that the `interface` can remain flexible and unopinionated on how the `interface` is implemented. 

An option for adding some sort of CORS option to the `interface` is to make it generic enough to work with any 'server implementation'. This requires a bit of research.

But if we add `CorsOption` as a parameter to the express router package (ie this package) we are able to provide the CORS ability to express router while keeping the interface free of any express details.

FYI: I did a lot of the code here: https://github.com/awslabs/fhir-works-on-aws-routing/pull/18/files

src/router/metadata/cap.rest.security.template.ts
AllanHodkinson: The FHIR spec says that CORS _should_ be implemented although it's not compulsory:
https://www.hl7.org/fhir/http.html#3.1.0.1.4
I can't think of a situation where you wouldn't want it available.

rsmayda: Good find on the link - when looking through it I found: 

> Note: to support browser-based client applications, servers SHOULD implement cross-origin resource sharing (CORS)  for the interactions documented here.

So I guess one use-case is if the FHIR server doesn't have a need to support browser-based client applications, then setting up CORS is not needed

datarster (dismissed):

rsmayda (commented):
Code looks good! 

Looks like we may need to add some changes to Severless to fully support CORS:
1. https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html
> For a Lambda proxy integration or HTTP proxy integration, you can still set up the required OPTIONS response headers in API Gateway.

2. https://www.serverless.com/blog/cors-api-gateway-survival-guide
> CORS with custom authorizers: "if you reject an authorization request, you don't have the ability to specify the CORS headers in your response. This can make it difficult for the client browser to understand the response."

Adding `cors: true` and that `GatewayResponseDefault4XX` seem to be the right approach

let me know your thoughts -- also when you get a chance can you describe how you tested this?

AllanHodkinson:
Hi Robert,
Yes this will require an update to the serverless model to support this:
an OPTIONS method for the `{proxy+}` route without authorization which will pass the request to the Lambda handler.

As you know, we have our own Cloudformation template based environment for deployment where we have added the equivalent method to the API gateway.  I've deployed the updated code under this environment into a new API gateway stage/endpoint and tested the following:
* A pre-flight request from Postman returns a 204 response with the expected CORS headers.
* A GET request from Postman returns a 200 response with the expected CORS headers.
* Setting a custom CORS configuration for `allowedHeaders` (using a CorsOptions TS structure passed to `generateServerlessRouter`) propagates through to the middleware, and is visible in a pre-flight OPTIONS Postman request.
* An Angular web app running in Chrome is able to make calls to the same API gateway endpoint.  Network tab shows that CORS requests are being served correctly.

The rejected authorization problem is an interesting one - we've actually seen this as an issue with a malformed auth token.  The solution in that blog will work, it's not strictly correct as you may have specific header values rather than wildcards, but is probably a pragmatic solution.

rsmayda (approved):
Okay great thanks for confirming your test cases -- Would you also be willing to add a blurb about these required serverless changes if CORS is enabled to the [CUSTOMIZE.md](https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/CUSTOMIZE.md#api-customization) file?

If not no biggie we will make the changes. Once I hear back from you I will merge these in!

Thanks for contributing back 👍 🚀

PR #19, opened 2020-10-29
feat: use new FWoA interface

mgresko:
Description of changes:

Depends on:  implementation of interface here https://github.com/awslabs/fhir-works-on-aws-interface/pull/22

### DIFF
--- a/src/router/metadata/cap.rest.security.template.ts
+++ b/src/router/metadata/cap.rest.security.template.ts
@@ -20,7 +20,8 @@ export default function makeSecurity(authConfig: Auth) {
                 },
             ],
         };
-        if (authConfig.strategy.oauthUrl) {
+        if (authConfig.strategy.oauth) {
+
             security = {
                 ...security,
                 ...{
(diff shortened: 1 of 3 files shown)


src/router/metadata/cap.rest.security.template.ts
```diff
@@ -20,7 +20,7 @@ export default function makeSecurity(authConfig: Auth) {
                 },
             ],
         };
-        if (authConfig.strategy.oauthUrl) {
+        if (authConfig.strategy.oauth?.oauthAuthorizationUrl) {
```
rsmayda: For this line we just need to ensure that the oauth object is present - if it is then we know `oauthAuthorizationUrl` will be there as it is required
```suggestion
        if (authConfig.strategy.oauth) {

```

rsmayda (commented):
Commits have to follow semantic-commits

So for you it could be
`feat: use new FWoA interface`

Also this PR will be on hold until we merge in the new interface and update the NPM (so that the unit tests pass)

PR #20, opened 2020-11-04
feat: implement 3.0 interface changes

rsmayda:
Description of changes:
- implemented the 3.0 interface
  - this encompasses changes from: https://github.com/awslabs/fhir-works-on-aws-routing/pull/19/files
- Made the operation outcome more explicit from a severity and code perspective

### DIFF
--- a/src/router/routes/errorHandling.ts
+++ b/src/router/routes/errorHandling.ts
@@ -1,6 +1,13 @@
 import express from 'express';
 import createError from 'http-errors';
-import { InvalidResourceError, ResourceNotFoundError, ResourceVersionNotFoundError } from 'fhir-works-on-aws-interface';
+import {
+    isInvalidResourceError,
+    isResourceNotFoundError,
+    isResourceVersionNotFoundError,
+    isUnauthorizedError,
+    IssueSeverity,
+    IssueCode,
+} from 'fhir-works-on-aws-interface';
 import OperationsGenerator from '../operationsGenerator';
 
 export const applicationErrorMapper = (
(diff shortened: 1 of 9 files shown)


rsmayda (commented):
@mgresko this PR encompasses your PR;; https://github.com/awslabs/fhir-works-on-aws-routing/pull/19

src/router/routes/errorHandling.ts
```diff
@@ -10,25 +17,39 @@ export const applicationErrorMapper = (
...
+};
+
 export const httpErrorHandler = (err: any, req: express.Request, res: express.Response, next: express.NextFunction) => {
     if (createError.isHttpError(err)) {
         console.error('HttpError', err);
-        res.status(err.statusCode).send(OperationsGenerator.generateError(err.message));
+        const { severity, code } = statusToOutcome[err.statusCode];
+        res.status(err.statusCode).send(
+            OperationsGenerator.generateOperationOutcomeIssue(severity ?? 'error', code ?? 'processing', err.message),
+        );
```
carvantes: L49 will crash if the error code is not in the map. I think that the map covers all the status codes that we are using now but this could break later on. A crash on the error handler is really bad since there's no outer layer to handle it.

I agree with the intent of having a default severity/code of 'error'/'processing', just refactor so that it does not crash.

carvantes (changes requested):

src/router/routes/errorHandling.ts
rsmayda: Fixed and tested via [playground](https://www.typescriptlang.org/play?ts=4.0.2#code/MYewdgzgLgBNCGUCuEAqIDySqgLYFMAuGAJX1ACcATAHjCVwCN8KAaGAbznwDcWBLKAE9i0CvzABzANwxQVInCjipMAL4A+GAF5OAKBiGYAFgAMp4lwi8Bw4gHIWFEBXvt5i+xJ7wANvyp7dVYDIzMAZktuPnE7GEcKZ1d3EAUHa2AkWKEgtRCjE1NjKOsYwRF4pxc3OVTPMBAoAFoAMxAkMEDg0MMAVnMSm2yHKuTatMqAD2B8AAcofnBckLVpPT1QSFgnHX0ChGQIAGE6wgAmcz01dc3oTmjbIRSFdV2DlHQsHBACAG0nAB072OdQAujAAPwQ+6lR4jRLVZ6eWbOGYQCASSS5G7gCAgXz4AG+ECSAAUsOySIAlEA)

carvantes (approved):

PR #24, opened 2020-11-11
feat: Merging 2.0.0 into mainline

nguyen102:
Issue #, if available:

Description of changes:
Merging 2.0.0 into mainline

This code has been tested with interface: 3.0.0 and passes Crucible integration tests and Bulk Export Integration tests.

Previously approved PRs:
https://github.com/awslabs/fhir-works-on-aws-routing/pull/23
https://github.com/awslabs/fhir-works-on-aws-routing/pull/20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 - Support for DB export by routing export requests to the corresponding BulkDataAccess interfaces as defined in `fhir-works-on-aws-interface` v3.0.0
 - Supporting capability statement configuration for OAuth as defined in `fhir-works-on-aws-interface` v3.0.0
 - Improved error handling to allow matching of same error objects across different `fhir-works-on-aws-interface` versions 
+- Support for configuring CORs header
 
 ## [1.1.0] - 2020-09-25
 - feat: Pass down allowed resource types to search service
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.0.0] - 2020-11-11
+
+### Added
+- Support for DB export by routing export requests to the corresponding BulkDataAccess interfaces as defined in `fhir-works-on-aws-interface` v3.0.0
+- Supporting capability statement configuration for OAuth as defined in `fhir-works-on-aws-interface` v3.0.0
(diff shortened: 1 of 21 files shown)


carvantes (dismissed):

carvantes (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.0.0] - 2020-11-11
+
+### Added
+- Support for DB export by routing export requests to the corresponding BulkDataAccess interfaces as defined in `fhir-works-on-aws-interface` v3.0.0
+- Supporting capability statement configuration for OAuth as defined in `fhir-works-on-aws-interface` v3.0.0
+- Improved error handling to allow matching of same error objects across different `fhir-works-on-aws-interface` versions 
+
```
carvantes: Don't forget to mention the CORS support on the changelog. It's not part of this merge but it is part of the 2.0.0 release.

Btw, I know that we haven't standarized our release process, but I don't like modifying the changelog on the same commit that is adding the feature. I think that is cleaner to do a separate commit specifically for release stuff (changelog, version bump, etc). That creates a commit history that is easier to understand.

carvantes (dismissed):

src/app.ts
```diff
@@ -53,24 +54,31 @@ export function generateServerlessRouter(
...
-                next();
-            } else {
-                res.status(403).json({ message: 'Forbidden' });
-            }
+            res.locals.requesterUserId = fhirConfig.auth.authorization.getRequesterUserId(accessToken);
+            next();
         } catch (e) {
-            res.status(403).json({ message: `Forbidden. ${e.message}` });
+            next(e);
         }
```
rsmayda, line 65: Do we have to do this try/catch? wouldn't our error mapper catch it?

rsmayda (dismissed):

src/app.ts
nguyen102: Yep, the try catch is needed. I was testing the code without the try catch with crucible. Crucible tried to do a GET on an invalid resource, example `/Foo/1`. Auth threw an error but the error was not caught by the error mapper. The reason is because we need to pass the error object to next. Passing the error object to next allows the next interceptor to receive the error object

CHANGELOG.md
nguyen102: updated

rsmayda (approved):

PR #25, opened 2020-11-17
feat: use new interface

rsmayda:
Description of changes:

- Implement new interface: https://github.com/awslabs/fhir-works-on-aws-interface/blob/mainline/src/authorization.ts

NOTE: will be in draft until new interface is released


nguyen102 (approved):
Looks good. Do we need to update the package version though?

rsmayda:
> Looks good. Do we need to update the package version though?

It will be a separate PR for Changelog & version bump

PR #28, opened 2020-11-21
fix: metadata and well-known endpoint should not have any auth

emilhdiaz:
Description of changes:

FHIR and SMART-on-FHIR require that the /metadata and /.well-known/smart-configuration endpoint NOT be behind auth. 

This change shifts the registration of these routes in the express router above the authentication middleware hook, so they are evaluated before auth kicks in. I think it makes more sense to make this change here than add conditional logic in the concrete implementations of `Authorization.verifyAccessToken()` to skip access token verification. 

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -51,6 +51,17 @@ export function generateServerlessRouter(
         hasCORSEnabled = true;
     }
 
+    // Metadata
+    const metadataRoute: MetadataRoute = new MetadataRoute(fhirVersion, configHandler, hasCORSEnabled);
+    app.use('/metadata', metadataRoute.router);
+
+    // well-known URI http://www.hl7.org/fhir/smart-app-launch/conformance/index.html#using-well-known
+    const smartStrat: SmartStrategy = fhirConfig.auth.strategy.oauthPolicy as SmartStrategy;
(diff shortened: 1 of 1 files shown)


nguyen102:
Thanks for the PR Emil. This change makes sense, let me test the code and I'll get back to you soon.

nguyen102:
I tested the changes in this PR to confirm that `metadata` was still reachable without Auth. I also ran Crucible tests, and those also passed. 

Thanks for thee contribution @emilhdiaz, I'll merge this PR in along with the Auth PR once it's been approved.
https://github.com/awslabs/fhir-works-on-aws-authz-rbac/pull/18

nguyen102 (approved):

emilhdiaz:
Thanks @nguyen102 ! Happy to contribute. Should we make a similar change in the `SMARTHandler`? 

https://github.com/awslabs/fhir-works-on-aws-authz-smart/blob/mainline/src/smartHandler.ts#L56

nguyen102:
Thanks for the suggestion Emil. I've created a PR for it here.
https://github.com/awslabs/fhir-works-on-aws-authz-smart/pull/6

PR #32, opened 2020-12-09
fix: authorize requester has permission to view all resources returned in the Bundle

nguyen102:
Description of changes:
For Bundle requests, filter out resources that the requester is not allowed to see. 

For example:
1. A Bundle contains entries for a WRITE request and a READ request
2. The read request fails with an Unauthorized error, because the requester doesn't have permission to see that resource
3. We will return the whole Bundle, but with a 403 entry for the READ response. An example bundle response is shown below

Example below
```
{
   "resourceType":"Bundle",
   "id":"b5fedc68-359a-4a7a-a2a3-b3cb1d86e0b7",
   "type":"transaction-response",
   "link":[
      {
         "relation":"self",
         "url":"https://API_URL.com/"
      }
   ],
   "entry":[
      {
         "response":{
            "status":"201 Created",
            "location":"Patient/e8baed5d-4a64-45ea-ae93-77d1796fc4d6",
            "etag":"1",
            "lastModified":"2020-12-09T22:28:37.613Z"
         }
      },
      {
         "resource":{
            
         },
         "response":{
            "status":"403 Forbidden",
            "location":"Patient/47135b80-b721-430b-9d4b-1557edc64947",
            "etag":"1",
            "lastModified":"expect.stringMatching(utcTimeRegExp)"
         }
      }
   ]
}

```

### DIFF
--- a/src/router/bundle/bundleGenerator.ts
+++ b/src/router/bundle/bundleGenerator.ts
@@ -6,6 +6,7 @@
 import uuidv4 from 'uuid/v4';
 import URL from 'url';
 import { SearchResult, BatchReadWriteResponse } from 'fhir-works-on-aws-interface';
+import { isEmpty } from 'lodash';
 
 type LinkType = 'self' | 'previous' | 'next' | 'first' | 'last';
 
@@ -101,7 +102,15 @@ export default class BundleGenerator {
 
         const entries: any = [];
         bundleEntryResponses.forEach(bundleEntryResponse => {
-            const status = bundleEntryResponse.operation === 'create' ? '201 Created' : '200 OK';
+            let status = '200 OK';
+            if (bundleEntryResponse.operation === 'create') {
+                status = '201 Created';
+            } else if (
+                ['read', 'vread'].includes(bundleEntryResponse.operation) &&
+                isEmpty(bundleEntryResponse.resource)
+            ) {
+                status = '403 Forbidden';
+            }
             const entry: any = {
                 response: {
                     status,
--- a/src/router/bundle/bundleHandler.ts
+++ b/src/router/bundle/bundleHandler.ts
@@ -23,7 +23,7 @@ import BundleHandlerInterface from './bundleHandlerInterface';
 import BundleGenerator from './bundleGenerator';
 import BundleParser from './bundleParser';
 
-// Using import will not import the library therefore using the syntax below to pull in the library
+// Using import will not import the library correctly, therefore using this syntax to pull in the library instead
 // https://github.com/es-shims/Promise.allSettled/issues/5
 const allSettled = require('promise.allsettled');
 
--- a/src/router/bundle/bundleHandler.ts
+++ b/src/router/bundle/bundleHandler.ts
@@ -23,10 +23,6 @@ import BundleHandlerInterface from './bundleHandlerInterface';
 import BundleGenerator from './bundleGenerator';
 import BundleParser from './bundleParser';
 
-// Using import will not import the library correctly, therefore using this syntax to pull in the library instead
-// https://github.com/es-shims/Promise.allSettled/issues/5
-const allSettled = require('promise.allsettled');
-
 export default class BundleHandler implements BundleHandlerInterface {
     private bundleService: Bundle;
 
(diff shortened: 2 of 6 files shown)


tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "esModuleInterop": true, // Enables emit interoperability between CommonJS and ES Modules via creation of namespace objects for all imports. Implies 'allowSyntheticDefaultImports'.
     "forceConsistentCasingInFileNames": true, // Disallow inconsistently-cased references to the same file.
     "outDir": "lib",
+    "lib": ["ES2020.Promise", "dom"],
```
nguyen102, line 9: Allow support for `Promise.allSettled`. Once we added a library we needed to add "dom" also. If we don't add "dom" we get the error " Cannot find name 'URL'. Did you mean 'url'" when running `yarn install`

"dom" adds support for the library URL.

rsmayda: thanks for the explainer

src/router/bundle/bundleGenerator.ts
```diff
@@ -101,7 +102,15 @@ export default class BundleGenerator {
 
         const entries: any = [];
         bundleEntryResponses.forEach(bundleEntryResponse => {
-            const status = bundleEntryResponse.operation === 'create' ? '201 Created' : '200 OK';
+            let status = '200 OK';
+            if (bundleEntryResponse.operation === 'create') {
+                status = '201 Created';
+            } else if (
+                ['read', 'vread'].includes(bundleEntryResponse.operation) &&
+                isEmpty(bundleEntryResponse.resource)
```
rsmayda, line 110: Will `bundleEntryResponse.resource` be empty for the scenario for reading a Patient that isn't there? (ie not found)

src/router/bundle/bundleHandler.ts
```diff
@@ -146,6 +147,32 @@ export default class BundleHandler implements BundleHandlerInterface {
             }
         }
 
+        const authAndFilterReadPromises = requests.map((request, index) => {
+            if (['read', 'vread', 'history-type', 'history-instance', 'search-type'].includes(request.operation)) {
```
rsmayda: Can you do `system-search` in bundles?

src/router/bundle/bundleHandler.ts
```diff
@@ -146,6 +147,32 @@ export default class BundleHandler implements BundleHandlerInterface {
             }
         }
 
+        const authAndFilterReadPromises = requests.map((request, index) => {
```
rsmayda, line 160: are these promises? You might need to add `async`

src/router/bundle/bundleHandler.ts
```diff
@@ -146,6 +147,32 @@ export default class BundleHandler implements BundleHandlerInterface {
...
+                });
+            }
+            return Promise.resolve();
+        });
+
+        const readResponses = await Promise.allSettled(authAndFilterReadPromises);
+
+        requests.forEach((request, index) => {
+            const responseEntry = bundleServiceResponse.batchReadWriteResponses[index];
+            if (['read', 'vread', 'history-type', 'history-instance', 'search-type'].includes(request.operation)) {
```
rsmayda: Do we need to include system reads?

src/router/bundle/bundleGenerator.ts
nguyen102: That case can not happen because an error is thrown in the `persistence` layer when it tries to read a resource that does not exist. We would throw a `ResourceNotFoundError` which would eventually propagate to the `ErrorHandler` in the router package. The whole Bundle transaction would fail and any changes made would be rolled back.

src/router/bundle/bundleHandler.ts
nguyen102: https://flaviocopes.com/javascript-async-await-array-map/
"Remember, we must wrap any code that calls await in an async function."

Yep, they are promises, but since we don't initiate `await` inside of the map function, we don't have to prefix the function with `async`

src/router/bundle/bundleHandler.ts
nguyen102: Technically you can, but in our implementation we only support transactions that contains CRUD requests. So in this particular use case we would only ever be filtering `read` responses. However, I'll add `search-system`  and `history-system` to the list to be complete.

https://github.com/awslabs/fhir-works-on-aws-routing/blob/496350f499776d5ebec6a6d1a601e7def26178a3/src/router/bundle/bundleParser.ts#L72

src/router/bundle/bundleHandler.ts
nguyen102: Are system reads, `search-system` and `history-system`? 
https://github.com/awslabs/fhir-works-on-aws-interface/blob/mainline/src/constants.ts#L70

In our implementation of Bundles we only support CRUD requests inside the Bundle, but I'll add `system-search` and `history-system` to this array for completeness.

src/router/bundle/bundleHandler.ts
rsmayda: Super helpful link thanks

rsmayda (approved):

PR #35, opened 2020-12-17
fix: When parsing Bundles for reference only fields explicitly named  'reference' should be considered a reference

nguyen102:
Issue #, if available:
FHIR-450

Description of changes:
fix: When parsing Bundles for reference only fields explicitly named  'reference' should be considered a reference

### DIFF
--- a/src/router/bundle/bundleParser.test.ts
+++ b/src/router/bundle/bundleParser.test.ts
@@ -3,7 +3,7 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 
-import { BatchReadWriteRequest, ResourceNotFoundError } from 'fhir-works-on-aws-interface';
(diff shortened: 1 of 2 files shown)


src/router/bundle/bundleParser.test.ts
```diff
@@ -1331,5 +1331,55 @@ describe('parseResource', () => {
                 );
             }
         });
+
+        test('Do not consider "referenceSeq" as a "reference". Only resource fields explicitly named "reference" should be considered a reference', async () => {
```
nguyen102, line 1335: Without the code change in `bundleParser.ts` this test will fail with this error
`Error: This entry's reference is not recognized. Entry's reference is: http://www.ncbi.nlm.nih.gov/nuccore . Valid format includes "<url>/resourceType/id" or "<urn:uuid:|urn:oid:><id>
`

With the change in `bundleParser.ts`, this test passes.

rsmayda (approved):

PR #52, opened 2021-02-05
refactor: remove extra messaging in InvalidResourceError error handler

nguyen102:
Issue #, if available:

Description of changes:
fix: If creating a new resource failed because of InvalidResource, propogate that error message to user as a 400 error

Related PR: 
https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/pull/51

### DIFF
--- a/src/router/handlers/resourceHandler.ts
+++ b/src/router/handlers/resourceHandler.ts
@@ -3,7 +3,16 @@
  *  SPDX-License-Identifier: Apache-2.0
  */
 
-import { Search, History, Persistence, FhirVersion, Authorization, KeyValueMap } from 'fhir-works-on-aws-interface';
+import {
+    Search,
+    History,
+    Persistence,
+    FhirVersion,
+    Authorization,
+    KeyValueMap,
+    isInvalidResourceError,
+} from 'fhir-works-on-aws-interface';
+import createError from 'http-errors';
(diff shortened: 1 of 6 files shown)


src/router/handlers/resourceHandler.ts
```diff
@@ -41,8 +50,15 @@ export default class ResourceHandler implements CrudHandlerInterface {
...
-        return createResponse.resource;
+        try {
+            const createResponse = await this.dataService.createResource({ resourceType, resource });
+            return createResponse.resource;
+        } catch (e) {
+            if (isInvalidResourceError(e)) {
+                throw new createError.BadRequest(e.message);
+            }
+            throw e;
+        }
```
carvantes: Isn't the error handler middleware already doing this?
https://github.com/awslabs/fhir-works-on-aws-routing/blob/mainline/src/router/routes/errorHandling.ts#L30

is this just to set a different error message? If so, It'd be better to remove the `Failed to parse request body as JSON` thing from the error handler middleware and move it to wherever that JSON validation error is being thrown. Then the error middleware can just use the incoming error message as is

nguyen102: updated

carvantes (approved):
looks good.

nit: I suggest renamong PR/squash commit to `refactor: remove extra messaging in InvalidResourceError error handler`

Issue #53, opened 2021-02-09
Create interaction without response causes runtime error


AllanHodkinson:
The return type of the persistence component `create` method is of type `GenericResponse`.  The `response` element is optional, if it's not present we get an undefined runtime error from this line of code:

https://github.com/awslabs/fhir-works-on-aws-routing/blob/8134be9422fce238c205532a78f915c26dc40834/src/router/routes/genericResourceRoute.ts#L175

Solution is to test for existence of `response` before testing for for existence of `response.meta`.
I _think_ that the Express response .send should work as expected if response is undefined.

nguyen102:
Thank you for bringing this up and the PR to fix it. Closing this GH Issue as this is now resolved.

PR #54, opened 2021-02-10
fix: GenericResponse response must exist if sub properties tested.

AllanHodkinson:
Issue #, if available:
https://github.com/awslabs/fhir-works-on-aws-routing/issues/53

Description of changes:
Add check that optional `response` exists before checking sub properties of it. 


src/router/routes/genericResourceRoute.ts
```diff
@@ -1,263 +1,263 @@
...
+                        userIdentity: res.locals.userIdentity,
+                        readResponse: response,
+                    });
+                    if (updatedReadResponse && updatedReadResponse.meta) {
+                        res.set({
+                            ETag: `W/"${updatedReadResponse.meta.versionId}"`,
+                            'Last-Modified': updatedReadResponse.meta.lastUpdated,
+                        });
+                    }
+                    res.send(updatedReadResponse);
```
carvantes, line 51: I agree that since we are using `any` for some return values, type safety cannot be guaranteed and this kind of additional checks are warranted. But, have this caused runtime errors on your side? I don't think that `this.authService.authorizeAndFilterReadResponse` can actually return null or undefined, even if the typing allows it (but I haven't tested extensively) 

The change on L45 only matters if `updatedReadResponse` is null or undefined and it prevents the code from crashing on L45. But then an undefined `updatedReadResponse` would be sent as response on L51 and that would break things elsewhere. This change looks good, some future work for us would be to better handle undefined responses or to narrow down our typescript types if undefined responses don't occur to begin with.

carvantes (approved):
Thanks for the PR.

The change looks good. Looks like your branch is now out-of-date with mainline. I can merge this after you sync your branch with manline.

src/router/routes/genericResourceRoute.ts
AllanHodkinson: We've only seen runtime errors for the 'create' operation case when not returning a response body.  I just filled in the cases for other operations returning a `GenericResponse` type without too much thought!

I agree that this is an obscure corner case - by design & the FHIR spec the persistence component `read` method should either return a resource or should throw a ResourceNotFoundError.  As you say future work might be to update the `read` interface to always require a resource.
  
In our use case we have different people implementing persistence components, with varying degrees of FHIR & developer knowledge, so even though it is wrong, there is a chance that one of them would actually fail to set a resource within the GenericResponse.  If we are using the stubbed `passThroughAuthz` component for `authorizeAndFilterReadResponse`, that would return the undefined `readResponse` into this code and trigger the bug.  I tested the code on L51 if `updatedReadResponse` is undefined and it is handled ok.  I expect this is dealt with somewhere in Express.

PR #55, opened 2021-02-10
feat: Use validators array from FhirConfig to validate resources

nguyen102:
Issue #, if available:

Description of changes:
FHIR-473

Use validators array from FhirConfig to validate resources

### DIFF
--- a/src/router/validation/ValidationUtilities.ts
+++ b/src/router/validation/ValidationUtilities.ts
@@ -1,0 +1,9 @@
+import { Validator } from 'fhir-works-on-aws-interface';
+
+export async function validateResource(validators: Validator[], resource: any) {
+    return Promise.all(
+        validators.map((validator: Validator) => {
+            return validator.validate(resource);
+        }),
+    );
+}
--- a/src/router/validation/ValidationUtilities.ts
+++ b/src/router/validation/ValidationUtilities.ts
@@ -1,9 +1,0 @@
-import { Validator } from 'fhir-works-on-aws-interface';
-
-export async function validateResource(validators: Validator[], resource: any) {
-    return Promise.all(
-        validators.map((validator: Validator) => {
-            return validator.validate(resource);
-        }),
-    );
-}
--- a/src/router/validation/validationUtilities.test.ts
+++ b/src/router/validation/validationUtilities.test.ts
@@ -1,0 +1,41 @@
+import { InvalidResourceError, Validator } from 'fhir-works-on-aws-interface';
+import { validateResource } from './validationUtilities';
+
+describe('validateResource', () => {
+    const mockedSuccessValidator: Validator = {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        validate(resource: any) {
+            return Promise.resolve();
+        },
+    };
+    const mockedFailedValidatorA: Validator = {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        async validate(resource: any) {
+            // Sleep for one second, so if validatorA and validatorB are in validators array we can ensure validatorA
+            // takes longer to run than validator B. This is helpful for testing that the validators are run sequentially
+            await new Promise(r => setTimeout(r, 1000));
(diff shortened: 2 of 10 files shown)


src/router/validation/ValidationUtilities.ts
```diff
@@ -0,0 +1,9 @@
+import { Validator } from 'fhir-works-on-aws-interface';
+
+export async function validateResource(validators: Validator[], resource: any) {
```
rsmayda: add the return type

src/router/validation/ValidationUtilities.ts
```diff
@@ -0,0 +1,9 @@
...
+    return Promise.all(
+        validators.map((validator: Validator) => {
```
rsmayda: can you use `validators.every` instead https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

src/router/validation/ValidationUtilities.ts
rsmayda: Could you add a unit test for this util with just some 'dummy' validators (ie always return true or false)

src/router/validation/ValidationUtilities.ts
```diff
@@ -0,0 +1,9 @@
+import { Validator } from 'fhir-works-on-aws-interface';
+
+export async function validateResource(validators: Validator[], resource: any) {
+    return Promise.all(
```
carvantes: We should run the validators in order. Otherwise the validation error for the same input may be non-deterministic.

carvantes (changes requested):

src/router/validation/ValidationUtilities.ts
nguyen102: added

src/router/validation/ValidationUtilities.ts
nguyen102: I'm going to change this to use a for loop that runs the validator sequentially as per @carvantes comment about deterministic error.

src/router/validation/ValidationUtilities.ts
nguyen102: Adding

src/router/validation/ValidationUtilities.ts
nguyen102: updated

src/router/validation/validationUtilities.ts
```diff
@@ -0,0 +1,9 @@
+import { Validator } from 'fhir-works-on-aws-interface';
+
+export async function validateResource(validators: Validator[], resource: any): Promise<void> {
+    for (let i = 0; i < validators.length; i += 1) {
+        // eslint-disable-next-line no-await-in-loop
+        await validators[i].validate(resource);
+    }
+    return Promise.resolve();
```
carvantes: This return statement is redundant

src/router/validation/validationUtilities.test.ts
```diff
@@ -0,0 +1,41 @@
...
+        validate(resource: any) {
+            return Promise.resolve();
+        },
+    };
+    const mockedFailedValidatorA: Validator = {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        async validate(resource: any) {
+            // Sleep for one second, so if validatorA and validatorB are in validators array we can ensure validatorA
+            // takes longer to run than validator B. This is helpful for testing that the validators are run sequentially
+            await new Promise(r => setTimeout(r, 1000));
```
carvantes: I like the idea here, but you only need a couple milliseconds to make your point. Idling the tests for 1 second on every run is a bit much.

src/router/validation/validationUtilities.ts
nguyen102: removed

src/router/validation/validationUtilities.test.ts
nguyen102: updated to 10 ms

carvantes (approved):

PR #58, opened 2021-02-18
fix: Correct cap statement to set updateCreate as false

nguyen102:
Issue #, if available:

Description of changes:
fix: Correct cap statement to set updateCreate as false

### DIFF
--- a/src/router/metadata/cap.rest.resource.template.ts
+++ b/src/router/metadata/cap.rest.resource.template.ts
@@ -55,7 +55,6 @@ export function makeGenericResources(
     const resources: any[] = [];
 
     const resourceOperations: any[] = makeOperation(operations);
-    const updateCreate: boolean = operations.includes('update');
     const hasTypeSearch: boolean = operations.includes('search-type');
 
     fhirResourcesToMake.forEach((resourceType: string) => {
@@ -63,7 +62,7 @@ export function makeGenericResources(
             makeResourceObject(
                 resourceType,
                 resourceOperations,
-                updateCreate,
+                false,
                 hasTypeSearch,
                 searchCapabilityStatement[resourceType],
             ),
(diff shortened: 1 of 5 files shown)


src/router/metadata/cap.rest.resource.template.ts
```diff
@@ -75,16 +74,9 @@ export function makeGenericResources(
...
     const capabilities: SearchCapabilityStatement = hasTypeSearch ? await resource.typeSearch.getCapabilities() : {};
 
-    return makeResourceObject(
-        resourceType,
-        resourceOperations,
-        updateCreate,
-        hasTypeSearch,
-        capabilities[resourceType],
-    );
+    return makeResourceObject(resourceType, resourceOperations, false, hasTypeSearch, capabilities[resourceType]);
```
carvantes: This is more accurate right now, but we should stop using hardcoded values.

We should be able to use the `updateCreateSupported` field from the `Persistence` interface to populate this https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/mainline/src/dataServices/dynamoDbDataService.ts#L62

nguyen102: updated

src/router/metadata/cap.rest.resource.template.ts
```diff
@@ -73,9 +73,8 @@ export function makeGenericResources(
     return resources;
 }
 
-export async function makeResource(resourceType: string, resource: Resource) {
+export async function makeResource(resourceType: string, resource: Resource, updateCreate: boolean) {
```
carvantes: The new parameter is not needed. This function can figure it out with `resource.persistence.updateCreateSupported `

rsmayda (approved):

PR #59, opened 2021-02-22
feat: Compile IG StructureDefinitions

emilhdiaz:
This change is part of a set of MRs across multiple repos for a proposal to add the ability to parse StructureDefinitions from IG packages and use the information found in those definitions to inject the supported FHIR profiles into the CapabilityStatement.


src/implementationGuides/index.ts
```diff
@@ -0,0 +1,58 @@
...
+     * @param resources - an array of FHIR resources. See: https://www.hl7.org/fhir/profiling.html
+     */
+    async compile(resources: any[]): Promise<any> {
+        const validStructureDefinitions: FhirStructureDefinition[] = [];
+        resources.forEach(s => {
+            if (StructureDefinitionImplementationGuides.isFhirStructureDefinition(s)) {
+                validStructureDefinitions.push(s);
+            } else {
+                throw new Error(`The following input is not a StructureDefinition: ${s.type} ${s.name}`);
+                // throw new Error(`The following input is not a StructureDefinition: ${JSON.stringify(s, null, 2)}`);
```
arthuston: Do you want to remove this?

arthuston (dismissed):

src/implementationGuides/index.ts
```diff
@@ -0,0 +1,58 @@
...
+    baseDefinition: string;
+    type: string;
+};
+
+/**
+ * This class compiles StructuredDefinitions from IG packages
+ */
+export class StructureDefinitionImplementationGuides implements ImplementationGuides {
+    /**
+     * Compiles the contents of an Implementation Guide into an internal representation used to build Elasticsearch queries.
```
carvantes: update/remove this comment

src/registry/index.ts
```diff
@@ -0,0 +1,50 @@
+import { FhirVersion, CapabilityRegistry, ResourceCapabilityStatement } from 'fhir-works-on-aws-interface';
+import { FhirStructureDefinition } from "../implementationGuides";
+
+/**
+ * This class is the single authority over the supported FHIR StructuredDefinition and their definitions
+ */
+export class FHIRStructureDefinitionRegistry implements CapabilityRegistry {
+    private readonly capabilityStatement: ResourceCapabilityStatement;
+
+    constructor(fhirVersion: FhirVersion, compiledImplementationGuides?: any[]) {
```
carvantes: looks like fhirVersion is unused and can be removed

emilhdiaz: Do we want to process and compile FHIR US CORE R3 & R4 as part of the default packaging? Similar to how it's done for SearchParameters? If so, then we could make use of the `fhirVersion` to implement the conditional loading of those default compiled files.

carvantes: We briefly considered that, but we decided that it is not a feature that we want to support right now.

It could simplify the setup for some customers, but we are not sure if the majority will find it useful. The thing is that for a given FHIR version there may be multiple versions of an IG. i.e. for FHIR 4.0.1 there is at least US Core 3.1.0 and 3.1.1 and we are not sure if everyone just want the latest version or if they prefer to use specific versions.

Also, we can't load US Core by default since not all customers want to use it. We'd need some kind of change to `FhirConfig` to specify which of the "precompiled" IGs should be loaded.

It is something that we could do on the future though.

emilhdiaz: Got it! I'll remove the `fhirVersion` param then.

src/implementationGuides/index.ts
emilhdiaz: Removed!

src/implementationGuides/index.ts
emilhdiaz: Updated!

src/registry/index.ts
emilhdiaz: Removed the `fhirVersion` param.

emilhdiaz:
This MR is blocked until the interface changes in https://github.com/awslabs/fhir-works-on-aws-interface/pull/55 are merged and a new version is generated and pinned here for the interface package.

carvantes:
Looks good. I just merged this PR into a feature branch. It'll be merged to mainline and released to npm on a later date alongside the updates to the deployment package.

Thanks again for contributing!

PR #63, opened 2021-03-04
feat: add HAPI FHIR Lambda Validator

carvantes:
Description of changes:

Add HAPI FHIR Lambda Validator. It's mostly a wrapper that calls a lambda function and parses the response.


src/router/validation/hapiFhirLambdaValidator.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    async validate(resource: any): Promise<void> {
+        const params = {
+            FunctionName: this.hapiValidatorLambdaArn,
+            InvocationType: 'RequestResponse',
+            Payload: JSON.stringify(resource),
+        };
+
+        const lambdaResponse = await this.lambdaClient.invoke(params).promise();
```
rsmayda, line 43: Should we have a timeout here or does the lambda itself have a short time-out?

src/router/validation/hapiFhirLambdaValidator.ts
```diff
@@ -0,0 +1,58 @@
...
+    severity: string;
+    msg: string;
+}
+
+interface HapiValidatorResponse {
+    errorMessages: ErrorMessage[];
+    successful: boolean;
+}
+
+export default class HapiFhirLambdaValidator implements Validator {
```
rsmayda, line 22: Just to confirm this is being added but not enabled by default?

carvantes: It is not enabled by default.

To enable it it needs to be added to FhirConfig: https://github.com/awslabs/fhir-works-on-aws-deployment/blob/mainline/src/config.ts#L75

nguyen102:
Overall looks good. Same question as Robert about the java lambda timeout. Should we add code to catch and log a timeout error?

src/router/validation/hapiFhirLambdaValidator.ts
carvantes: It's probably fine to not set a client timeout here and just use the Lambda timeout. It is currently 1 minute but I'll lower it.

It's worth noting that, even with provisioned concurrency, there can be spikes of call volume that trigger cold starts, so we need to have a relatively high timeout value. It's probably better to allow a few requests to have latencies of >10 seconds due to cold starts than to have those requests explicitly fail due to a client timeout.

carvantes:
> Overall looks good. Same question as Robert about the java lambda timeout. Should we add code to catch and log a timeout error?

If the Lambda function times out it'll result in an unhandled error, which is ok since that means that validation failed to run altogether. The default logging for unhandled errors is sufficient

nguyen102 (dismissed):

src/router/validation/hapiFhirLambdaValidator.ts
rsmayda: What would be the error returned to the client if the validation takes longer than 30 secs (API GW timeout)?

rsmayda:
> If the Lambda function times out it'll result in an unhandled error, which is ok since that means that validation failed to run altogether. The default logging for unhandled errors is sufficient


What is the experience from an end user perspective? With API GW time out at 30 secs what is returned? HTTP code and message?

src/router/validation/hapiFhirLambdaValidator.ts
carvantes: If API GW times out after 30 seconds. clients get:
```json
504 Gateway Timeout
{
    "message": "Endpoint request timed out"
}
```
Which is not so great since it is not wrapped in an `OperationOutcome` but this is existing behavour regardless of what the validator does.

I made a change to add a lambda client timeout. I'll still change the time out on CFN but it's ok to have the timeout in both places. If the lambda client times out, an exception is thrown which is not handled(there's nothing meaningful to do) and ends up as a 500 error with a proper `OperationOutcome` "Internal Server Error" response

src/router/validation/hapiFhirLambdaValidator.ts
```diff
@@ -24,14 +26,18 @@ export default class HapiFhirLambdaValidator implements Validator {
...
+            },
+        });
     }
 
     async validate(resource: any): Promise<void> {
         const params = {
             FunctionName: this.hapiValidatorLambdaArn,
             InvocationType: 'RequestResponse',
-            Payload: JSON.stringify(resource),
+            Payload: JSON.stringify(JSON.stringify(resource)),
```
rsmayda, line 40: Do you mean to stringify twice?

carvantes: yes. You almost always need to stringify at least once since `Payload` MUST be a string(in a sense, Lambda is doing `JSON.parse` once before passing the event to your function handler). 

In our case we want the payload to be the resource serialized as a JSON string, so we end up with 2 calls to `stringify`

rsmayda: ahh today I learned -- thanks!

rsmayda (approved):

PR #64, opened 2021-03-17
fix: Relax validation rules before a patch can be applied

chrisbednarski:
fhir spec specifies the supported patch content-types at https://www.hl7.org/fhir/http.html#patch

The structure of each patch type is dependent on the content-type.
`application/json-patch+json` patch type is being rejected due to very strict validation rules

### DIFF
--- a/src/router/routes/genericResourceRoute.ts
+++ b/src/router/routes/genericResourceRoute.ts
@@ -218,7 +218,7 @@ export default class GenericResourceRoute {
                     const { id } = req.params;
                     const { body } = req;
 
-                    if (body.id === null || body.id !== id) {
+                    if (body.id && body.id !== id) {
                         throw new createError.BadRequest(
                             `Can not update resource with ID[${id}], while the given request payload has an ID[${body.id}]`,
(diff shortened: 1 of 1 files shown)


lyndoh:
Seems to me that body.id check should be removed completely, since none of the patch formats have a body.id field ?

chrisbednarski:
happy to update the pr, if required

rsmayda:
Hey thanks for cutting us a PR, looks like a good commit! Looks like our Semantic review failed. I updated the title of the PR to follow our standard.

When you get the chance could you commit an empty commit to retrigger the check

https://coderwall.com/p/vkdekq/git-commit-allow-empty

chrisbednarski:
all checks pass now

rsmayda (approved):

rsmayda:
This will be added to the next release. Do you have any timelines you are working against with this change? The next release should be near the end of March 2021

chrisbednarski:
Great. Thanks for accepting my PR. End of March suits us fine.

PR #65, opened 2021-03-25
fix: fix stu3 schema for Bundle & AllergyIntolerance

carvantes:
Turns out that the [JSON Schema files from the STU3 FHIR website](http://hl7.org/fhir/STU3/validation.html#json-schema) have some errors in them.  

**AllergyIntolerance**
Bad declaration of array items type

**Bundle**
It used an ambiguous `oneOf` since it included things like `DomainResource` that don't have a `resourceType` property that serves as a discriminator.

```
{
  "keyword": "oneOf",
  "dataPath": ".entry[0].resource",
  "schemaPath": "#/oneOf",
  "params": {
    "passingSchemas": [
      38,
      80
    ]
  },
  "message": "should match exactly one schema in oneOf"
}
```

The above fixes are present on the R4 schema.

#### Related Issues:
- https://github.com/awslabs/fhir-works-on-aws-deployment/issues/261


src/router/validation/schemas/fhir.schema.v3.json
```diff
@@ -4793,15 +4793,14 @@
             },
             "category": {
               "description": "Category of the identified substance.",
-              "enum": [
```
rsmayda, line 4796: Are there other enums in STU3 that are like this? if this is a large task we can put it in our backlog

rsmayda (approved):

nguyen102 (approved):

src/router/validation/schemas/fhir.schema.v3.json
carvantes: there may be more, It's hard to spot them just by looking. Scripting can aid in finding out. I'll add an item to the backlog.

PR #70, opened 2021-04-08
feat: Post search

Bingjiling:
Issue #, if available:

Description of changes:

* Add POST for search
* Tested with local deployment 
* Updated file mode to 755 to be in sync with other FHIR packages
* Behavior of the parameters merge by example:

| parametersInQuery        | parametersInBody | mergedParameterResult |
|--------------------------|------------------|-----------------------|
| active=true              | active=false     | active: [true, false] |
| active=true              | active=true      | active: true          |
| active=true              | None             | active: true          |
| active=true&active=false | active=false     | active: [true, false] | 


src/router/routes/genericResourceRoute.ts
```diff
@@ -135,23 +155,36 @@ export default class GenericResourceRoute {
...
+                    if (isEmpty(conflictSearchParams)) {
+                        const updatedReadResponse = await handleSearch(res, resourceType, {
+                            ...searchParamQuery,
+                            ...body,
+                        });
+                        res.send(updatedReadResponse);
+                    } else {
+                        throw new createError.BadRequest(
+                            `Can not process search request, query parameter and request body parameter conflicts in [${conflictSearchParams}]`,
+                        );
```
carvantes: duplicate params are actually ok.

https://www.hl7.org/fhir/http.html#search
> This has exactly the same semantics as the equivalent GET command. Note that in the POST variant, parameters may appear in both the URL and the body. Parameters have the same meaning in either place. Since parameters can repeat, putting them in both places is the same as repeating them (which is valid for some parameters and not for others).

I think it's better for the the router to not validate any search parameters. It can just forward them to the search component and it's up to search to decide if repeated parameters make sense.

Granted, this is a bit hard to test since currently fwoa crashes when search parameters are repeated...
The only search params that are properly handled when repeated are `_include` and `_revinclude` (and their `:iterate` versions)
i.e.
```
{{API_URL}}/MedicationRequest?_include=MedicationRequest:subject&_include=MedicationRequest:performer&_revinclude=MedicationAdministration:request&_revinclude=Provenance:target
```

src/router/routes/genericResourceRoute.ts
```diff
@@ -135,23 +155,36 @@ export default class GenericResourceRoute {
...
-                        allowedResourceTypes,
-                        res.locals.userIdentity,
-                    );
-                    const updatedReadResponse = await this.authService.authorizeAndFilterReadResponse({
-                        operation: 'search-type',
-                        userIdentity: res.locals.userIdentity,
-                        readResponse: response,
+                    const conflictSearchParams: string[] = [];
+                    forOwn(searchParamQuery, (value, key) => {
+                        if (has(body, key) && body[key] !== value) {
```
carvantes: `value` may be an array. express adds duplicate query params into an array

`url/?a=1&?a=2&b=3` results in 
```
req = {
  query: {
    a: [1, 2],
    b: 3
  }
}
```

carvantes (changes requested):
Overall looks good, but some changes are needed in how to handle duplicate params

src/router/routes/genericResourceRoute.ts
```diff
@@ -135,23 +155,36 @@ export default class GenericResourceRoute {
                     const resourceType = req.baseUrl.substr(1);
                     const searchParamQuery = req.query;
 
-                    const allowedResourceTypes = await this.authService.getAllowedResourceTypesForOperation({
-                        operation: 'search-type',
-                        userIdentity: res.locals.userIdentity,
-                    });
+                    const updatedReadResponse = await handleSearch(res, resourceType, searchParamQuery);
```
rsmayda:
```suggestion
                    const searchResponse = await handleSearch(res, resourceType, searchParamQuery);
```

src/router/routes/genericResourceRoute.ts
```diff
@@ -127,6 +128,25 @@ export default class GenericResourceRoute {
...
+                    userIdentity: res.locals.userIdentity,
+                });
+
+                const response = await this.handler.typeSearch(
+                    resourceType,
+                    searchParamQuery,
+                    allowedResourceTypes,
+                    res.locals.userIdentity,
+                );
+                const updatedReadResponse = await this.authService.authorizeAndFilterReadResponse({
```
rsmayda:
```suggestion
                const updatedSearchResponse = await this.authService.authorizeAndFilterReadResponse({
```

rsmayda (commented):
Could you also plan to update the integration tests in the deployment? this is a good example of how to: https://github.com/awslabs/fhir-works-on-aws-deployment/pull/284

src/router/routes/genericResourceRoute.ts
Bingjiling: Thanks for the detailed explanation Nestor! Just so I understand the expected behavior, for the above example you gave, the parameters passed to the search component would be:  

```
{
  _includ: [MedicationRequest:subject, MedicationRequest:performer],
  _revinclude: [MedicationAdministration:request, Provenance:target]
}
```

And in the case where repeated parameters don't make sense (e.g. active=true&active=false), routing should still pass `active:[true, false]` for search component to decided. 

Is that the right way to understand it?

carvantes: that's correct

carvantes:
> Updated file mode to 755 to be in sync with other FHIR packages

Why? by default 755 is only used for folders

e.g.
```
186590e08e87:fhir-works-on-aws-search-es nestorba$ stat -f '%A %N' *
644 CHANGELOG.md
644 CODE_OF_CONDUCT.md
644 CONTRIBUTING.md
644 LICENSE
644 NOTICE
644 README.md
644 THIRD-PARTY
644 jest.config.js
755 lib
755 node_modules
644 package.json
755 scripts
755 src
644 tsconfig.json
644 yalc.lock
644 yarn.lock
```

Bingjiling:
> > Updated file mode to 755 to be in sync with other FHIR packages
> 
> Why? by default 755 is only used for folders
> 
> e.g.
> 
> ```
> 186590e08e87:fhir-works-on-aws-search-es nestorba$ stat -f '%A %N' *
> 644 CHANGELOG.md
> 644 CODE_OF_CONDUCT.md
> 644 CONTRIBUTING.md
> 644 LICENSE
> 644 NOTICE
> 644 README.md
> 644 THIRD-PARTY
> 644 jest.config.js
> 755 lib
> 755 node_modules
> 644 package.json
> 755 scripts
> 755 src
> 644 tsconfig.json
> 644 yalc.lock
> 644 yarn.lock
> ```

Ah! My mistake, I'll change it back.

src/router/routes/genericResourceRoute.ts
```diff
@@ -127,31 +128,56 @@ export default class GenericResourceRoute {
...
+                    const searchParamQuery = req.query;
+                    const { body } = req;
 
-                    const allowedResourceTypes = await this.authService.getAllowedResourceTypesForOperation({
-                        operation: 'search-type',
-                        userIdentity: res.locals.userIdentity,
-                    });
+                    if (!isEmpty(body)) {
+                        mergeWith(searchParamQuery, body, (valueOne, valueTwo) => {
+                            const result: string | any[] = [];
```
carvantes: `result` is never of type `string`. Furthermore, this var is not really needed, see my other comments

src/router/routes/genericResourceRoute.ts
```diff
@@ -127,31 +128,56 @@ export default class GenericResourceRoute {
...
+                            if (!isEmpty(valueOne) && !isEmpty(valueTwo) && !(valueOne === valueTwo)) {
+                                return [...new Set(result.concat(valueOne).concat(valueTwo))];
```
carvantes: concat accepts N params
```suggestion
                                return [...new Set([].concat(valueOne, valueTwo))];
```

src/router/routes/genericResourceRoute.ts
```diff
@@ -127,31 +128,56 @@ export default class GenericResourceRoute {
...
+                    const { body } = req;
 
-                    const allowedResourceTypes = await this.authService.getAllowedResourceTypesForOperation({
-                        operation: 'search-type',
-                        userIdentity: res.locals.userIdentity,
-                    });
+                    if (!isEmpty(body)) {
+                        mergeWith(searchParamQuery, body, (valueOne, valueTwo) => {
+                            const result: string | any[] = [];
+                            if (!isEmpty(valueOne) && !isEmpty(valueTwo) && !(valueOne === valueTwo)) {
```
carvantes: There is the case when both GET and POST params are arrays:
```
parametersInQuery=active=true&active=false
parametersInBody=active=true&active=false
```

This code works in that case but it is misleading since it does `valueOne === valueTwo` and equality comparisons between arrays is always false.

I suggest to simply drop the `!(valueOne === valueTwo)` condition. Duplicated params are removed anyways by the Set

Bingjiling: If `!(valueOne === valueTwo)` is dropped, then the situation where `parametersInQuery=active=true` and `parametersInBody=active=true` would result in [true] instead of true.

carvantes: Right.

Still, let's not do `===` comparisons on arrays. We can be explicit about merging arrays OR unique values by replacing `!(valueOne === valueTwo)` with something like `(Array.isArray(valueOne) || Array.isArray(valueTwo) || valueOne !== valueTwo)`

btw, deduplicating params wasn't strictly necessary, although it doesn't hurt to have it here. Maybe that logic fits better in the Search component since it ends up being only an optimization on the resulting search queries.

Bingjiling: `(Array.isArray(valueOne) || Array.isArray(valueTwo) || valueOne !== valueTwo)` won't work for situation like `parametersInQuery=active=true` and `parametersInBody=gender=female`. It'll result in `active=[true, undefined]`, `gender=[undefined,female]`. We'll still need the block `(!isEmpty(valueOne) || !isEmpty(valueTwo)` to return undefined before it. 

I agree it's not ideal to use `===` on arrays, but the alternative is getting the conditions more complicated. I can add a comment here explaining what situation it covers. Does that work?

carvantes: sorry, I meant replacing the last part of the condition, not all of it. 
i.e.
```
!isEmpty(valueOne) &&
!isEmpty(valueTwo) &&
(Array.isArray(valueOne) || Array.isArray(valueTwo) || valueOne !== valueTwo)
```
yes, the conditions are more complex, but I think that they are easier to understand as a developer (95% of the time when you see an array `==` comparison it is a bug). It's kind of adding more code to make your intent clear instead of adding a comment to explain it.

Bingjiling: Okay, I'll update the condition to avoid `===` on arrays.

carvantes (approved):

PR #73, opened 2021-04-28
Create requestContext and pass it along with userIdentity

justinusmenzel:
Issue #, if available:

Description of changes:
- create requestContext based on info in express.Request
- store in express.Response.locals
- pass it through a VerifyAccessTokenRequest instance to fhirConfig.auth.authorization.verifyAccessToken
- pass it down through various functions next to userIdentity

depends on: https://github.com/awslabs/fhir-works-on-aws-interface/pull/62


src/app.ts
```diff
@@ -25,6 +27,23 @@ import { FHIRStructureDefinitionRegistry } from './registry';
 
 const configVersionSupported: ConfigVersion = 1;
 
+function prepareRequestContext(req: express.Request): RequestContext {
```
justinusmenzel, line 30: Should this function be here? I didn't find a good helper TS file. I don't want to put it in the interface package because it depends on express and interface doesn't seem to depend on anything.

rsmayda: hmm yeah I think you are right -- a util.ts file would be ideal but I don't seen an issue adding the method here

src/app.ts
```diff
@@ -25,6 +27,23 @@ import { FHIRStructureDefinitionRegistry } from './registry';
...
+        url: req.url,
+        contextInfo: {},
+    };
+
+    if (req.method) {
+        const method = req.method.toUpperCase();
+        if (['CONNECT', 'DELETE', 'GET', 'HEAD', 'OPTIONS', 'PATCH', 'POST', 'PUT', 'TRACE'].includes(method)) {
+            requestContext.verb = method as VerbType;
+        }
+    }
```
rsmayda, line 43: Anyway this `if` doesn't pass? can `req.method` be `null`? Same question for the types of methods?

src/router/handlers/exportHandler.ts
```diff
@@ -9,7 +9,7 @@ import {
     BulkDataAccess,
     GetExportStatusResponse,
     InitiateExportRequest,
-    KeyValueMap,
+    KeyValueMap, RequestContext,
```
rsmayda: nit: new line
```suggestion
    KeyValueMap, 
    RequestContext,
```

src/router/handlers/rootHandler.ts
```diff
@@ -22,9 +22,10 @@ export default class RootHandler {
         this.serverUrl = serverUrl;
     }
 
-    async globalSearch(queryParams: any, userIdentity: KeyValueMap) {
+    async globalSearch(queryParams: any, requestContext: RequestContext, userIdentity: KeyValueMap) {
```
rsmayda:
```suggestion
    async globalSearch(queryParams: any, userIdentity: KeyValueMap, requestContext: RequestContext) {

```

src/router/handlers/rootHandler.ts
```diff
@@ -35,9 +36,10 @@ export default class RootHandler {
         return BundleGenerator.generateBundle(this.serverUrl, queryParams, searchResponse.result, 'searchset');
     }
 
-    async globalHistory(queryParams: any, userIdentity: KeyValueMap) {
+    async globalHistory(queryParams: any, requestContext: RequestContext, userIdentity: KeyValueMap) {
```
rsmayda:
```suggestion
    async globalHistory(queryParams: any, userIdentity: KeyValueMap, requestContext: RequestContext) {

```

src/app.ts
justinusmenzel: Yes, technically the interface allows the `verb` to be undefined. That's because `req.method`'s type is a string but for `verb` only the union type values are allowed. In real life `req.method` should always have a valid value. In any case the auth code could deny a request without `verb` in the RequestContext.

src/router/handlers/exportHandler.ts
justinusmenzel: fixed.

src/router/handlers/rootHandler.ts
justinusmenzel: fixed.

src/router/handlers/rootHandler.ts
justinusmenzel: fixed.

justinusmenzel:
Since https://github.com/awslabs/fhir-works-on-aws-interface/pull/62 is merged I could update the package.json but we probably need a tagged version of interface?

rsmayda:
> need a tagged version of interface? 

Yep we need to 'release' that merge we hope to do that Tomorrow or Thursday. We will ping here when we release it!

rsmayda (approved):
Going to have a 2nd set of eyes looking at this as well

carvantes (approved):

nguyen102 (approved):

PR #75, opened 2021-04-30
fix: bad substitution in absolute URL refs that match the server url

carvantes:
Issue #, if available:
See: https://github.com/awslabs/fhir-works-on-aws-deployment/issues/305

Description of changes:
The resource type in references was wrongly being set to the resourceType of the resource that contains the reference.


src/router/bundle/bundleParser.test.ts
```diff
@@ -513,7 +513,7 @@ describe('parseResource', () => {
                             ],
                         },
                         subject: {
-                            reference: 'Observation/47135b80-b721-430b-9d4b-1557edc64947',
```
carvantes, line 516: Existing test was actually asserting the wrong behavior

src/regExpressions.ts
```diff
@@ -22,5 +22,5 @@ export const captureVersionIdRegExp = /\w+\/[-\w+]+\/_history\/(\w+)/;
...
...lProduct|MedicinalProductAuthorization|MedicinalProductContraindication|MedicinalProductIndication|MedicinalProductIngredient|MedicinalProductInteraction|MedicinalProductManufactured|MedicinalProductPackaged|MedicinalProductPharmaceutical|MedicinalProductUndesirableEffect|MessageDefinition|MessageHeader|MolecularSequence|NamingSystem|NutritionOrder|Observation|ObservationDefinition|OperationDefinition|OperationOutcome|Organization|OrganizationAffiliation|Patient|PaymentNotice|PaymentReconciliation|Person|PlanDefinition|Practitioner|PractitionerRole|Procedure|Provenance|Questionnaire|QuestionnaireResponse|RelatedPerson|RequestGroup|ResearchDefinition|ResearchElementDefinition|ResearchStudy|ResearchSubject|RiskAssessment|RiskEvidenceSynthesis|Schedule|SearchParameter|ServiceRequest|Slot|Specimen|SpecimenDefinition|StructureDefinition|StructureMap|Subscription|Substance|SubstanceNucleicAcid|SubstancePolymer|SubstanceProtein|SubstanceReferenceInformation|SubstanceSourceMaterial|SubstanceSpecification|SupplyDelivery|SupplyRequest|Task|TerminologyCapabilities|TestReport|TestScript|ValueSet|VerificationResult|VisionPrescription)\/([A-Za-z0-9\-.]{1,64})(\/_history\/[A-Za-z0-9\-.]{1,64})?/;
```
rsmayda, line 25: Looks like the delta is right in the beginning? 
`(?:http|https)`
what is the `?:` doing in the beginning of the regex?

nguyen102 (dismissed):

src/regExpressions.ts
carvantes: `?:` signals a non-capturing group.

It's a good practice to use non-capturing groups if you don't actually care about the captured values. In this particular case they are useful to prevent "shifting" of the capture values(the code that uses this regex already assumes that certain values are on a certain index in the captured values array)

https://www.regular-expressions.info/brackets.html

rsmayda (approved):

PR #81, opened 2021-05-12
feat: Add winston logger

Bingjiling:
Issue #, if available:

Description of changes:

* Add Winston logger to routing package, tested by deploying to an AWS account and checking the log printed


src/loggerBuilder.ts
```diff
@@ -0,0 +1,8 @@
+import { makeLogger } from 'fhir-works-on-aws-interface';
+
+export default function makeComponentLogger(metadata?: any): any {
+    return makeLogger({
+        component: 'routing',
+        ...metadata,
+    });
+}
```
carvantes: This will result in each file having its own logger instance. Ideally the entire package should be using a single logger instance.

Maybe try exporting the logger object instead.
```js
export const logger =  makeLogger({
        component: 'routing',
        ...metadata,
    });
}
```

looks like the routing package has only log statement so it's not a big deal,  but applying this same pattern on other packages will result in many loggers.

carvantes (changes requested):

src/loggerBuilder.ts
Bingjiling: Sure, I can update it so one package share one logger. I started this way thinking we might want to have metadata on the class level, but end up not adding class level metadata.

carvantes (approved):
lgtm

PR #83, opened 2021-05-13
feat: enable $docref based on the compiled IGs

carvantes:
The idea is to support `docref` but at the same time we want to make it easier to add other operations in the future.
 
Summary of changes:

* Update the compiler to also accept `OperationDefinition`
* Added `OperationDefinitionRegistry` to hold the implementation of available operations.
* Update `app.ts` to hook operations into our authZ middleware and also mount their implementations.

To add another operation in the future we only need to code it’s implementation (add a folder under `/operationDefinitions`) and  add it on  `initializeOperationRegistry` . 

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -22,6 +22,7 @@ import { applicationErrorMapper, httpErrorHandler, unknownErrorHandler } from '.
 import ExportRoute from './router/routes/exportRoute';
 import WellKnownUriRouteRoute from './router/routes/wellKnownUriRoute';
 import { FHIRStructureDefinitionRegistry } from './registry';
+import { initializeOperationRegistry } from './operationDefinitions';
 
 const configVersionSupported: ConfigVersion = 1;
 
@@ -38,6 +39,7 @@ export function generateServerlessRouter(
     const serverUrl: string = fhirConfig.server.url;
     let hasCORSEnabled: boolean = false;
     const registry = new FHIRStructureDefinitionRegistry(compiledImplementationGuides);
+    const operationRegistry = initializeOperationRegistry(configHandler);
 
     const app = express();
     app.use(express.urlencoded({ extended: true }));
@@ -70,7 +72,9 @@ export function generateServerlessRouter(
     // AuthZ
     app.use(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
         try {
-            const requestInformation = getRequestInformation(req.method, req.path);
+            const requestInformation =
+                operationRegistry.getOperation(req.method, req.path)?.requestInformation ??
+                getRequestInformation(req.method, req.path);
             // Clean auth header (remove 'Bearer ')
             req.headers.authorization = cleanAuthHeader(req.headers.authorization);
             res.locals.userIdentity = await fhirConfig.auth.authorization.verifyAccessToken({
@@ -92,6 +96,11 @@ export function generateServerlessRouter(
         );
         app.use('/', exportRoute.router);
     }
+
+    // Operations defined by OperationDefinition resources
+    operationRegistry.getAllRouters().forEach(router => {
+        app.use('/', router);
+    });
 
     // Special Resources
     if (fhirConfig.profile.resources) {
--- a/src/configHandler.ts
+++ b/src/configHandler.ts
@@ -4,6 +4,7 @@
  */
 
 import { FhirConfig, FhirVersion, TypeOperation } from 'fhir-works-on-aws-interface';
+import ResourceHandler from './router/handlers/resourceHandler';
 
 export default class ConfigHandler {
     readonly config: FhirConfig;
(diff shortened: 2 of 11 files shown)


src/app.ts
```diff
@@ -38,6 +39,7 @@ export function generateServerlessRouter(
     const serverUrl: string = fhirConfig.server.url;
     let hasCORSEnabled: boolean = false;
     const registry = new FHIRStructureDefinitionRegistry(compiledImplementationGuides);
+    const operationRegistry = initializeOperationRegistry(configHandler);
```
rsmayda, line 42: This is probably too meta/another PR but could/should we use this for existing operations i.e. 'create', 'search', etc? Another way to put it is should all operations be treated this way?

src/configHandler.ts
```diff
@@ -66,4 +67,36 @@ export default class ConfigHandler {
 
         return resources;
     }
+
+    getResourceHandler(resourceType: string): ResourceHandler | undefined {
```
rsmayda: I think a small comment explaining the method will help

```suggestion
    // @returns specialized resourceHandler, if not found returns generic resourceHandler, if not valid returns undefined
    getResourceHandler(resourceType: string): ResourceHandler | undefined {

```

src/operationDefinitions/index.ts
```diff
@@ -0,0 +1,26 @@
...
+import { OperationDefinitionImplementation } from './types';
+import { USCoreDocRef } from './USCoreDocRef';
+import { OperationDefinitionRegistry } from './OperationDefinitionRegistry';
+import ConfigHandler from '../configHandler';
+
+export const initializeOperationRegistry = (configHandler: ConfigHandler) => {
+    const { compiledImplementationGuides } = configHandler.config.profile;
+    const operations: OperationDefinitionImplementation[] = [];
+
+    if (
```
rsmayda: Minor comment, thoughts?

```suggestion
   // Add US Core $docref operation if found in the compiled implementation guide. Note: this is an example and can be removed if not desired
    if (
```

src/operationDefinitions/types.ts
```diff
@@ -5,20 +5,36 @@
...
 
 export interface OperationDefinitionImplementation {
     /**
      * url of the corresponding OperationDefinition resource
      */
-    canonicalUrl: string;
+    readonly canonicalUrl: string;
+
+    /**
+     * url path used to invoke the operation
```
rsmayda, line 17: I think adding an example here will be helpful. Like do we expect the '/' to be there in the start or no?

```suggestion
     * url path used to invoke the operation. Example: "/$docref"
```

src/operationDefinitions/types.ts
```diff
@@ -5,20 +5,36 @@
...
+     */
+    readonly path: string;
+
+    /**
+     * methods supported by this operation e.g GET, POST
+     */
+    readonly httpMethods: string[];
```
rsmayda: In the Interface package we used the term 'verb' instead of 'method': https://github.com/awslabs/fhir-works-on-aws-interface/blob/mainline/src/authorization.ts#L23

src/operationDefinitions/types.ts
```diff
@@ -5,20 +5,36 @@
...
+
+    /**
+     * FHIR resourceType that is affected by this operation
+     */
+    readonly targetResourceType: string;
```
rsmayda, line 30: Are there situations where it is a global operation? could we add that layer in now or perhaps wait until that situation arises?

src/app.ts
carvantes: That's an interesting idea. At first glance I think that it's better not to mix the base operations from the RESTful API with the additional operations that come from `OperationDefinition`. They are treated/defined very differently in the FHIR spec. Although it'd be possible to encapsulate the RESTful operations in a similar way.

I don't see that as a high priority refactor though. IMO the base RESTful operations are (mostly) expected always be enabled, whereas `OperationDefinition` can be seen as extensions/plugins and benefit from this "more pluggable" approach.

https://www.hl7.org/fhir/operationdefinition.html#defined
https://www.hl7.org/fhir/http.html#3.1.0

src/configHandler.ts
carvantes: added comment

src/operationDefinitions/types.ts
carvantes: added the example

src/operationDefinitions/types.ts
carvantes: yes, there are global operations. Although the majority are resource-specific.

The main goal is to ship `$docref`, but we also want to provide guidance for implementing future operations. I think that stopping at an interface that works for resource-specific operations is a good compromise.

I think it's ok to keep it as is and update it later when we need to implement a global operation. This interface is private to routing so we can update it without too much trouble. 

https://www.hl7.org/fhir/operationdefinition.html#defined

src/operationDefinitions/types.ts
carvantes: I updated it for consistency with our interface (interface is a lot more visible), but I do think that `httpMethods` is a better name.

`method` is the official term in the HTTP spec. They are sometimes referred as `verb` but I don't recall seeing `verb` used in any api.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
https://datatracker.ietf.org/doc/html/rfc7231#section-4
http://expressjs.com/en/api.html#req.method
https://github.com/axios/axios#axios-api
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-method.html#cfn-apigateway-method-httpmethod

src/operationDefinitions/index.ts
carvantes: I added some comments here.
 
Yeah, I do think that customers are likely to modify this particular file to change how operations are enabled/disabled.

Right now customers that don't want to modify source code have the following options:
- Do not use US Core, $docref is disabled
- Add US Core IG, $docref is enabled
- Add US Core, but delete the $docref OperationDefinition file before compiling, $docref is disabled (this one is a bit awkward, but I believe that this is the least common use case)

rsmayda (approved):

PR #85, opened 2021-05-14
feat: add docref to capability statement

carvantes:
Add the operations registered in `OperationDefinitionRegistry` to the capability statement

I chose to add a `documentation` field on the implementation of the operation instead of just copying the `documentation` from the `OperationDefinition` from the IG since docs for operations are sometimes vague and it is expected that servers document the particularities of their implementation. e.g. our `docref` documents that it does not generate documents on demand.

The fragment of the cap statement for docref looks like this:
```
{
  "type": "DocumentReference",
  ...
  "supportedProfile": [
    "http://hl7.org/fhir/us/core/StructureDefinition/us-core-documentreference"
  ],
  "operation": [
    {
      "name": "docref",
      "definition": "http://hl7.org/fhir/us/core/OperationDefinition/docref",
      "documentation": "This operation is used to return all the references to documents related to a patient..."
    }
  ]
}
```

### DIFF
--- a/src/operationDefinitions/OperationDefinitionRegistry.test.ts
+++ b/src/operationDefinitions/OperationDefinitionRegistry.test.ts
@@ -13,6 +13,8 @@ import ResourceHandler from '../router/handlers/resourceHandler';
 const fakeRouter = (jest.fn() as unknown) as Router;
 const fakeOperation: OperationDefinitionImplementation = {
     canonicalUrl: 'https://fwoa.com/operation/fakeOperation',
+    name: 'fakeOperation',
+    documentation: 'The documentation for the fakeOperation',
     httpVerbs: ['GET'],
     path: '/Patient/fakeOperation',
     targetResourceType: 'Patient',
@@ -52,6 +54,31 @@ describe('OperationDefinitionRegistry', () => {
         expect(operationDefinitionRegistry.getOperation('GET', '/Patient/someOtherOperation')).toBeUndefined();
 
         expect(operationDefinitionRegistry.getOperation('GET', '/Patient/fakeOperation')).toBe(fakeOperation);
+    });
+
+    test('getCapabilities', () => {
+        const configHandlerMock = {
+            getResourceHandler: jest.fn().mockReturnValue({}),
+        };
+
+        const operationDefinitionRegistry = new OperationDefinitionRegistry(
+            (configHandlerMock as unknown) as ConfigHandler,
+            [fakeOperation],
+        );
+
+        expect(operationDefinitionRegistry.getCapabilities()).toMatchInlineSnapshot(`
+            Object {
+              "Patient": Object {
+                "operation": Array [
+                  Object {
+                    "definition": "https://fwoa.com/operation/fakeOperation",
+                    "documentation": "The documentation for the fakeOperation",
+                    "name": "fakeOperation",
+                  },
+                ],
+              },
+            }
+        `);
     });
 
     test('ResourceHandler not available', () => {
--- a/src/operationDefinitions/OperationDefinitionRegistry.ts
+++ b/src/operationDefinitions/OperationDefinitionRegistry.ts
@@ -8,6 +8,18 @@ import { Router } from 'express';
 import { OperationDefinitionImplementation } from './types';
 import ConfigHandler from '../configHandler';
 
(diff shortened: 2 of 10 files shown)


src/operationDefinitions/OperationDefinitionRegistry.test.ts
```diff
@@ -54,6 +56,31 @@ describe('OperationDefinitionRegistry', () => {
...
+            (configHandlerMock as unknown) as ConfigHandler,
+            [fakeOperation],
+        );
+
+        expect(operationDefinitionRegistry.getCapabilities()).toMatchInlineSnapshot(`
+            Object {
+              "Patient": Object {
+                "operation": Array [
+                  Object {
+                    "definition": "https://fwoa.com/operation/fakeOperation",
```
Bingjiling, line 74: Is definition always mapped to `canonicalUrl` ?

src/operationDefinitions/OperationDefinitionRegistry.ts
```diff
@@ -35,4 +47,23 @@ export class OperationDefinitionRegistry {
     getAllRouters(): Router[] {
         return this.routers;
     }
+
+    getCapabilities(): OperationCapabilityStatement {
+        const capabilities: OperationCapabilityStatement = {};
+
+        this.operations.forEach(operation => {
```
Bingjiling, line 54: Would `_.groupBy` be helpful here?

src/router/metadata/metadataHandler.test.ts
```diff
@@ -431,6 +446,45 @@ test('R4: FHIR Config V4 without search', async () => {
         updateCreate: configHandler.config.profile.genericResource!.persistence.updateCreateSupported,
     };
     expect(response.resource.rest[0].resource[0]).toMatchObject(expectedResourceSubset);
+    expect(response.resource.rest[0].resource[0]).toMatchInlineSnapshot(`
```
Bingjiling, line 443: What's the difference between `toMatchInlineSnapshot` and `toMatchObject` here except the formatting different? Looking at the Jest document, it only talks about using `toMatchInlineSnapshot` for front end testing.

src/operationDefinitions/OperationDefinitionRegistry.ts
carvantes: It could help, but it's not exactly what we need.

Notably there's an intemerdiate `opertaion` property that we need. `groupBy` would give `{ type: [o1,o2]}` but we want `{ type: { operation: [o1,o2]} }`

I could use `groupBy` and then rearrange the object, but i think is clearer to not use it in this case.

src/operationDefinitions/OperationDefinitionRegistry.test.ts
carvantes: yes

https://www.hl7.org/fhir/capabilitystatement-definitions.html#CapabilityStatement.rest.resource.operation.definition

src/router/metadata/metadataHandler.test.ts
carvantes: I updated to remove the `toMatchObject` expectation since it is redundant now.

`toMatchInlineSnapshot` matches the input against the string snapshot that appears below. It's mostly used in frontend due to it's ties with jest + react, but it can be used for anything that can be serialized into a string (i.e. JSON)

I added the snapshot to test that the `fakeOperation` is present. I could have updated the `expectedResourceSubset` instead but I find snapshots to be easier to understand in this case.
i.e., for `toMatchObject(expectedResourceSubset)` you need to see the definition of `expectedResourceSubset` and then the definition of the `makeOperation` function to understand the expected result. Snapshots immediately tell you the expected JSON. 

https://jestjs.io/docs/snapshot-testing#does-snapshot-testing-only-work-with-react-components

src/operationDefinitions/OperationDefinitionRegistry.ts
Bingjiling: Got it, that makes sense

src/router/metadata/metadataHandler.test.ts
Bingjiling: Emm that makes sense, good call

Bingjiling (approved):
LGTM

PR #94, opened 2021-06-08
feat: pass tenantId to subcomponents. Allow tenant specific URLs

carvantes:
Extract tenantId from the access token and pass it down to FWoA subcomponents. Also support tenant specific URLs such as
```
{{API_URL}}/tenant/tenant111/Patient
```

Notable changes:
- `src/router/middlewares/setTenantId.ts` extracts the tenantId from the access token and sets `res.locals.tenantId` for the rest of the application to use. 
The AuthZ components are unchanged, they only validate that the `tenantId` is legit by verifying the token signature.
- server url can not longer be statically set in class constructors, since it can differ for each tenant. `src/router/middlewares/setServerUrl.ts` resolves the `serverUrl` for the incoming request and sets in `res.locals` so that the rest of the application can reference it.

Most of the other code changes are just for passing `tenantId` to FWoA subcomponents.

**NOTE:** Unit tests won't pass since the interface changes are on a feature branch. They work fine locally.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -25,6 +25,8 @@ import ExportRoute from './router/routes/exportRoute';
 import WellKnownUriRouteRoute from './router/routes/wellKnownUriRoute';
 import { FHIRStructureDefinitionRegistry } from './registry';
 import { initializeOperationRegistry } from './operationDefinitions';
+import { setServerUrlMiddleware } from './router/middlewares/setServerUrl';
+import { setTenantIdMiddleware } from './router/middlewares/setTenantId';
 
 const configVersionSupported: ConfigVersion = 1;
 
@@ -63,7 +65,7 @@ export function generateServerlessRouter(
     const app = express();
     app.disable('x-powered-by');
 
-    const mainRouter = express.Router();
+    const mainRouter = express.Router({ mergeParams: true });
 
     mainRouter.use(express.urlencoded({ extended: true }));
     mainRouter.use(
@@ -78,6 +80,8 @@ export function generateServerlessRouter(
         mainRouter.use(cors(corsOptions));
         hasCORSEnabled = true;
     }
+
+    mainRouter.use(setServerUrlMiddleware(fhirConfig));
 
     // Metadata
     const metadataRoute: MetadataRoute = new MetadataRoute(
@@ -118,13 +122,13 @@ export function generateServerlessRouter(
         }
     });
 
+    if (fhirConfig.multiTenancyConfig?.enableMultiTenancy) {
+        mainRouter.use(setTenantIdMiddleware(fhirConfig));
+    }
+
     // Export
     if (fhirConfig.profile.bulkDataAccess) {
-        const exportRoute = new ExportRoute(
-            serverUrl,
-            fhirConfig.profile.bulkDataAccess,
-            fhirConfig.auth.authorization,
-        );
+        const exportRoute = new ExportRoute(fhirConfig.profile.bulkDataAccess, fhirConfig.auth.authorization);
         mainRouter.use('/', exportRoute.router);
     }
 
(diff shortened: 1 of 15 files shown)


src/app.ts
```diff
@@ -205,7 +209,11 @@ export function generateServerlessRouter(
     mainRouter.use(httpErrorHandler);
     mainRouter.use(unknownErrorHandler);
 
-    app.use('/', mainRouter);
+    if (fhirConfig.multiTenancyConfig?.enableMultiTenancy && fhirConfig.multiTenancyConfig?.useTenantSpecificUrl) {
+        app.use('/tenant/:tenantIdFromPath([a-zA-Z0-9\\-_]{1,64})', mainRouter);
```
rsmayda, line 213: How are the error cases handled? if I use an invalid URL ie url > 64 chars or url with a `+`?

Will Express just return a 404?

src/router/middlewares/setServerUrl.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Sets the value of `res.locals.serverUrl`
+ * the serverUrl can either be a static value from FhirConfig of a dynamic value for some multi-tenancy setups.
+ */
+export const setServerUrlMiddleware: (
+    fhirConfig: FhirConfig,
+) => (req: express.Request, res: express.Response, next: express.NextFunction) => void = (fhirConfig: FhirConfig) => {
+    return RouteHelper.wrapAsync(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+        if (req.baseUrl && req.baseUrl !== '/') {
+            res.locals.serverUrl = fhirConfig.server.url + req.baseUrl;
```
rsmayda, line 21: So `fhirConfig.server.url` is going to be the API_URL environment variable (`apigw.com/dev`) and `req.baseUrl` is the FHIR side (`/Patient/1234`). ie full url could be `apigw.com/dev/tenant/t1/Patient/1234`

Where would the tenant portion reside?

src/router/middlewares/setTenantId.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+export const setTenantIdMiddleware: (
+    fhirConfig: FhirConfig,
+) => (req: express.Request, res: express.Response, next: express.NextFunction) => void = (fhirConfig: FhirConfig) => {
+    return RouteHelper.wrapAsync(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
+        const tenantId = get(res.locals.userIdentity, fhirConfig.multiTenancyConfig?.tenantIdClaimPath!);
+        if (
+            tenantId === undefined ||
+            !tenantIdRegex.test(tenantId) ||
+            (req.params.tenantIdFromPath !== undefined && req.params.tenantIdFromPath !== tenantId)
```
rsmayda, line 26: Do we want to use OR here instead of AND?

if `tenantIdFromPath` is undefined it will equate to FALSE and not throw an unauthorized exception, is that what we want?

rsmayda: Looking at your tests it looks like this is covered? Maybe I am misunderstanding this if block

carvantes: This translates to: "If you have a tenantId in your URL, then it must match the tenantId in the JWT token"

`tenantIdFromPath` can be undefined if customers are not using tenant specific urls. That is fine

src/app.ts
carvantes: yes, it will be a 404

src/router/middlewares/setServerUrl.ts
carvantes: > req.baseUrl is the FHIR side (/Patient/1234)

That's incorrect, `/Patient/1234` would be `req.path`. `req.baseUrl` is the the URL path on which the router instance was mounted.

In our case the main router is either mounte on `/` or `tenant/:tenantId` depending in the value of `useTenantSpecificUrl`

rsmayda (approved):

PR #97, opened 2021-06-22
feat: pass tenantId to bundle service

carvantes:
Forgot to pass down tenantId to bundleService on previous PR

### DIFF
--- a/src/router/bundle/bundleHandler.ts
+++ b/src/router/bundle/bundleHandler.ts
@@ -58,8 +58,8 @@ export default class BundleHandler implements BundleHandlerInterface {
         this.validators = validators;
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    async processBatch(bundleRequestJson: any, userIdentity: KeyValueMap, requestContext: RequestContext) {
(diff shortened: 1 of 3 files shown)


src/router/bundle/bundleHandler.ts
```diff
@@ -58,8 +58,8 @@ export default class BundleHandler implements BundleHandlerInterface {
         this.validators = validators;
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    async processBatch(bundleRequestJson: any, userIdentity: KeyValueMap, requestContext: RequestContext) {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars,prettier/prettier
```
rsmayda, line 61: what are we disabling here?

rsmayda (approved):

src/router/bundle/bundleHandler.ts
carvantes: Now that we added an extra param, prettier wants the parameters to be one per line but that would require adding an extra `@typescript-eslint/no-unused-vars` on each line.

IMO It looks better to just disable prettier in this case.

PR #98, opened 2021-06-30
feat: extract tenantId from aud claim

Bingjiling:
Issue #, if available:

Description of changes:

* Extract tenantId from aud claim 
* Update/Add unit tests

### DIFF
--- a/src/router/middlewares/setTenantId.ts
+++ b/src/router/middlewares/setTenantId.ts
@@ -11,9 +11,9 @@ import RouteHelper from '../routes/routeHelper';
 
 const tenantIdRegex = /^[a-zA-Z0-9\-_]{1,64}$/;
 
-const getTenantIdFromAudClaim = (audClaim: string, baseUrl: string) => {
-    // aud claim should ALWAYS be present
-    if (!audClaim) {
+const getTenantIdFromAudClaim = (audClaim: any, baseUrl: string) => {
+    // aud claim could be an array for some IDP setups, check if aud claim is a string for compatibility
+    if (!audClaim || !(typeof audClaim === 'string')) {
         return undefined;
     }
     if (audClaim.startsWith(`${baseUrl}/tenant/`)) {
--- a/src/router/middlewares/setTenantId.ts
+++ b/src/router/middlewares/setTenantId.ts
@@ -11,16 +11,18 @@ import RouteHelper from '../routes/routeHelper';
 
 const tenantIdRegex = /^[a-zA-Z0-9\-_]{1,64}$/;
 
-const getTenantIdFromAudClaim = (audClaim: string) => {
-    // aud claim should ALWAYS be present regardless of tenancy mode
+const getTenantIdFromAudClaim = (audClaim: string, baseUrl: string) => {
+    // aud claim should ALWAYS be present
     if (!audClaim) {
         throw new UnauthorizedError('Unauthorized');
     }
-    const audClaimSplit = audClaim.split('://')[1].split('/');
-    if (audClaimSplit.length === 2) {
+    if (audClaim === baseUrl) {
         return undefined;
     }
-    return audClaimSplit.pop();
+    if (audClaim.startsWith(`${baseUrl}/tenant/`)) {
+        return audClaim.substring(`${baseUrl}/tenant/`.length);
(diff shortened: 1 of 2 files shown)


src/router/middlewares/setTenantId.ts
```diff
@@ -11,6 +11,18 @@ import RouteHelper from '../routes/routeHelper';
...
+    // aud claim should ALWAYS be present regardless of tenancy mode
+    if (!audClaim) {
+        throw new UnauthorizedError('Unauthorized');
+    }
+    const audClaimSplit = audClaim.split('://')[1].split('/');
+    if (audClaimSplit.length === 2) {
+        return undefined;
+    }
+    return audClaimSplit.pop();
+};
```
carvantes: Looks like this string parsing is incorrect. It allows invalid aud values like: 
- `https://xxxx.execute-api.us-east-2.amazonaws.com/dev/t2`
- `https://xxxx.execute-api.us-east-2.amazonaws.com/dev/many/other/things`

We need to match **exactly** `<baseUrl>/tenant/<tenantId>` (note that the above example is missing `/tenant`)

We should use the value of `res.locals.serverUrl` as `baseUrl` to do the string matching.

https://github.com/awslabs/fhir-works-on-aws-routing/blob/f29c54e9956db73cca22b68809e4580efb320d33/src/router/middlewares/setServerUrl.ts#L21

carvantes (changes requested):

src/router/middlewares/setTenantId.ts
```diff
@@ -11,6 +11,20 @@ import RouteHelper from '../routes/routeHelper';
...
+        throw new UnauthorizedError('Unauthorized');
+    }
+    if (audClaim === baseUrl) {
+        return undefined;
+    }
+    if (audClaim.startsWith(`${baseUrl}/tenant/`)) {
+        return audClaim.substring(`${baseUrl}/tenant/`.length);
+    }
+    throw new UnauthorizedError('Unauthorized');
+};
```
carvantes: I think that this method can be changed to something like:
```js
const getTenantIdFromAudClaim = (audClaim: string, baseUrl: string) => { 
    if (!audClaim) {
        return undefined;
    }
    if (audClaim.startsWith(`${baseUrl}/tenant/`)) {
        return audClaim.substring(`${baseUrl}/tenant/`.length);
    }
    return undefined;
};
```
In a sense it is a best effort to extract the tenantId from the `aud` claim.

Checking the existence and the validity of the `aud` claim is done on the [SMART package](https://github.com/awslabs/fhir-works-on-aws-authz-smart/blob/c644cb7a867d456cebb587a1f3cf6d036b6226af/src/smartAuthorizationHelper.ts#L140) or by the APIGW + Cognito intergration, so there's no need to redo it here.

Also, the cognito token doesn't use urls for `aud`. Cognito will always use custom claims for the tenantId and the router should not fail if the `aud` claim doesn't follow the `<baseUrl>/tenant/<tenantId>` schema.

```js
{
  "sub": "5ecb250e-8964-4120-ae25-xxxxxxxxx,
  "cognito:groups": [
    "practitioner"
  ],
  "email_verified": true,
  "iss": "https://cognito-idp.us-west-2.amazonaws.com/us-west-2_xxxxxxxx",
  "cognito:username": "workshopxxxxx",
  "origin_jti": "7467505b-e196-431d-b069-xxxxxxxxxx",
  "custom:tenantId": "tenant1",
  "aud": "2qi5m0hlnp4sxxxxxxxxxxxx",
  "event_id": "54f405a6-7e15-415e-b4f7-xxxxxxxxxxx",
  "token_use": "id",
  "auth_time": 1625138033,
  "exp": 1625141633,
  "iat": 1625138033,
  "jti": "07ef9e6a-b823-401b-bae1-xxxxxxxxxxx",
  "email": "dummy@email.com"
}
```

Sorry if all of this was not clear on the design doc.

carvantes: Also, the `aud` claim can be a string or an array of strings.

I'm not sure if our Okta implementation ever uses an array of strings for `aud`, but we should handle both cases for compatibility with other IDP setups.

carvantes (approved):

PR #99, opened 2021-07-13
fix: use correct server url and allow aud arrays

carvantes:
- Using `res.locals.serverUrl` instead of `fhirConfig.server.url` was a bug that caused to append `/tenant` twice and was messing up the tenantId extraction.
- Allow `aud` to be an array. Our SMART Okta implementation always has a single aud value, but other customers may have `aud` arrays


rsmayda (approved):
FYI: I believe the SMART package handles aud as an array or a string

PR #100, opened 2021-07-15
feat: pass fhirServiceBaseUrl as param to authZ calls

carvantes:
Description of changes:

Just passing down `res.locals.serverUrl` as param to the different AuthZ calls. 

See: https://github.com/awslabs/fhir-works-on-aws-interface/pull/72

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -115,6 +115,7 @@ export function generateServerlessRouter(
                 ...requestInformation,
                 requestContext: res.locals.requestContext,
                 accessToken: req.headers.authorization,
+                fhirServiceBaseUrl: res.locals.serverUrl,
             });
             next();
         } catch (e) {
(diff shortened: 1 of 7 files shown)


src/app.ts
```diff
@@ -115,6 +115,7 @@ export function generateServerlessRouter(
                 ...requestInformation,
                 requestContext: res.locals.requestContext,
                 accessToken: req.headers.authorization,
+                fhirServiceBaseUrl: res.locals.serverUrl,
```
rsmayda, line 118: This was probably set in a previous PR but where is `res.locals.serverUrl` initialized?

carvantes: Here:
https://github.com/awslabs/fhir-works-on-aws-routing/blob/552a1180697a6406950c4a27e6b0cc0b3a0b2f76/src/router/middlewares/setServerUrl.ts#L12-L16

rsmayda: Thanks

rsmayda (approved):

PR #101, opened 2021-07-22
fix: properly handle versioned references in BundleParser

justinusmenzel:
Issue #, if available:

Description of changes:
If a bundle contains versioned references to resources already in persistent storage, the checkReferences currently sends the string `/_history/<vid>` as vid to the vReadResource. This fix removes the `/_history/` prefix.

### DIFF
--- a/src/router/bundle/bundleParser.test.ts
+++ b/src/router/bundle/bundleParser.test.ts
@@ -790,7 +790,8 @@ describe('parseResource', () => {
                                 reference: `${serverUrl}/Patient/111`,
                             },
                             device: {
-                                reference: `${serverUrl}/Device/222`,
+                                // make sure versioned references also work
+                                reference: `${serverUrl}/Device/222/_history/7`,
                             },
                             effectiveDateTime: '2021-01-09T20:00:06Z',
                             valueQuantity: {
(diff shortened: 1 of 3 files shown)


src/router/bundle/bundleParser.test.ts
```diff
@@ -790,7 +790,8 @@ describe('parseResource', () => {
                                 reference: `${serverUrl}/Patient/111`,
                             },
                             device: {
-                                reference: `${serverUrl}/Device/222`,
+                                // make sure versioned references also work
+                                reference: `${serverUrl}/Device/222/_history/7`,
```
rsmayda, line 794: Just to understand this testcase. This is POSTing a bundle with a historical reference. The output of calling this method should have that reference with `_history/7`, right?

Maybe I am not understanding this testcase

justinusmenzel: I suppose it should. 
The main goal of this change was to avoid calling vReadResource with a vid that's not formatted correctly. 
But obviously we also want to store the reference with the version info. Maybe I don't understand the question correctly.

rsmayda (dismissed):
It looks like you are out of date could you rebase?

carvantes (dismissed):

carvantes (approved):

PR #102, opened 2021-07-22
fix: update FHIR v4 schema and fix the validator

jn1119:
Issue #, if available: [Issue-377](https://github.com/awslabs/fhir-works-on-aws-deployment/issues/377)

Description of changes: 
* Updated v4 schema from https://hl7.org/FHIR/fhir.schema.json.zip
* Fixed the validator to use `id` field if it was missing

Testing
* No new unit tests added. Existing unit tests already capture both the cases (with and without id)
* Also tested that API calls worked locally

### DIFF
--- a/src/router/validation/jsonSchemaValidator.ts
+++ b/src/router/validation/jsonSchemaValidator.ts
@@ -16,16 +16,21 @@ import fhirV3Schema from './schemas/fhir.schema.v3.json';
 export default class JsonSchemaValidator implements Validator {
     private ajv: any;
 
+    private readonly schemaId: string;
+
     constructor(fhirVersion: FhirVersion) {
         const ajv = new Ajv({ schemaId: 'auto', allErrors: true });
+        let schema;
         if (fhirVersion === '4.0.1') {
             ajv.addMetaSchema(schemaDraft06);
             ajv.addSchema(fhirV4Schema);
-        }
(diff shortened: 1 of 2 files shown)


nguyen102 (dismissed):

src/router/validation/jsonSchemaValidator.ts
```diff
@@ -16,15 +16,24 @@ import fhirV3Schema from './schemas/fhir.schema.v3.json';
...
+        } else if (fhirVersion === '3.0.1') {
             ajv.addMetaSchema(schemaDraft04);
             ajv.addSchema(fhirV3Schema);
+            schema = fhirV3Schema;
+        }
+        if (schema && 'id' in schema) {
+            this.schemaId = schema.id;
+        } else {
+            this.schemaId = ``;
         }
```
rsmayda: nit: Can simplify to single line ternary and use chaining: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining
```suggestion
        this.schemaId = schema?.id ?? '';
```

jn1119: I changed that statement into a ternary. The above didn't work for me because we need to also incorporate `id in schema` condition.

rsmayda (approved):

PR #107, opened 2021-07-27
feat: Group export map additional parameters needed

Bingjiling:
Issue #, if available:

Description of changes:

* Map additional parameters needed by group export
* allowedResourceTypes for filtering down the resource types
* serverUrl for validating if url reference points to resource on the same server
* fhirVersion for getting patientCompartment definition in the specific fhir version
* Depends on https://github.com/awslabs/fhir-works-on-aws-interface/pull/73 
* Map allowed resource types
* Fix broken unit tests 

### DIFF
--- a/src/router/bundle/bundleHandler.test.ts
+++ b/src/router/bundle/bundleHandler.test.ts
@@ -54,6 +54,9 @@ const genericResource: GenericResource = {
     typeHistory: stubs.history,
     typeSearch: stubs.search,
 };
+
+const dummyServerUrl: string = 'https://dummy-server-url';
+
 const resources = {};
 
 const SUPPORTED_R4_RESOURCES = [
@@ -414,7 +417,7 @@ describe('ERROR Cases: Validation of Bundle request', () => {
         bundleRequestJSON.type = 'batch';
 
         await expect(
-            bundleHandlerR4.processBatch(bundleRequestJSON, practitionerDecoded, dummyRequestContext),
+            bundleHandlerR4.processBatch(bundleRequestJSON, practitionerDecoded, dummyRequestContext, dummyServerUrl),
         ).rejects.toThrowError(new createError.BadRequest('Currently this server only support transaction Bundles'));
     });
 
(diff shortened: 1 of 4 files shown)


ssvegaraju (approved):

src/router/bundle/bundleHandler.test.ts
```diff
@@ -54,6 +54,9 @@ const genericResource: GenericResource = {
     typeHistory: stubs.history,
     typeSearch: stubs.search,
 };
+
+const dummyServerUrl: string = 'https://dummy-server-url';
```
rsmayda, line 58: How are these broken?

Bingjiling: These are broken before this code change. It's fixed in this PR but not relevant to the change here.

src/router/routes/exportRoute.ts
```diff
@@ -46,11 +66,15 @@ export default class ExportRoute {
...
+            '/Group/:id/\\$export',
+            RouteHelper.wrapAsync(async (req: express.Request, res: express.Response) => {
+                const exportType: ExportType = 'group';
+                await this.initiateExportRequests(req, res, exportType);
+            }),
+        );
 
-        this.router.get('/Group/:id/\\$export', () => {
+        this.router.get('/Patient/\\$export', () => {
             throw new createHttpError.BadRequest('We currently do not support Group export');
```
rsmayda:
```suggestion
            throw new createHttpError.BadRequest('We currently do not support Patient export');
```

src/router/routes/exportRoute.ts
```diff
@@ -16,18 +22,32 @@ export default class ExportRoute {
...
         this.exportHandler = new ExportHandler(bulkDataAccess, authService);
         this.init();
     }
 
     async initiateExportRequests(req: express.Request, res: express.Response, exportType: ExportType) {
+        const allowedResourceTypes = await this.authService.getAllowedResourceTypesForOperation({
+            operation: 'read',
+            userIdentity: res.locals.userIdentity,
+            requestContext: res.locals.userIdentity,
+        });
```
rsmayda, line 42: we will be calling this for system exports too -- will be changing it's existing functionality -- from if admin be able to export everything to only export based on what scopes you provided; or in RBAC what resources align with said group

I think this will make the export more aligned with AuthZ but we will need a major version bump here and clear README explaining the change

src/router/bundle/bundleHandler.test.ts
rsmayda: talked offline this is a feature branch

src/router/routes/exportRoute.ts
Bingjiling: Gotcha. Will update the commit message on merge to: 

```
feat!: add support for group data export

BREAKING CHANGE: System export will only include resources based on what scopes you provided(SMART) or what resources align with said group(RBAC)
```

rsmayda (approved):

carvantes (approved):

Issue #106, opened 2021-07-27
Validation Errors on CapabilityStatement


nathan-abacus:
**Describe the bug**
We use 'metadataRoute' to retrieve CapabilityStatement. The CapabilityStatement has these validation errors: 

`Validation of response body against profile 'http://hl7.org/fhir/StructureDefinition/CapabilityStatement' by FHIR specification's Validation Engine produced the following results:
ERROR: The Coding references a value set, not a code system ("http://hl7.org/fhir/ValueSet/restful-security-service"). Location: CapabilityStatement.rest[0].security.service[0].coding[0] (line 1, col 1039).
ERROR: Unknown Code System http://hl7.org/fhir/ValueSet/restful-security-service. Location: CapabilityStatement.rest[0].security.service[0].coding[0] (line 1, col 1039).
ERROR: Search parameter names must be unique in the context of a resource. [searchParam.select(name).isDistinct()]. Location: CapabilityStatement.rest[0].resource[41] (line 1, col 350391).
ERROR: Search parameter names must be unique in the context of a resource. [searchParam.select(name).isDistinct()]. Location: CapabilityStatement.rest[0].resource[43] (line 1, col 371006).
ERROR: Search parameter names must be unique in the context of a resource. [searchParam.select(name).isDistinct()]. Location: CapabilityStatement.rest[0].resource[102] (line 1, col 841601).`

**Screenshots**
<img width="621" alt="Screen Shot 2021-07-27 at 3 56 03 PM" src="https://user-images.githubusercontent.com/79218407/127219229-f2a69ac7-bfe0-4700-82b8-056f2625eb1c.png">

carvantes:
Hi,

Thanks for bringing this up.

We just fixed some of the errors on https://github.com/awslabs/fhir-works-on-aws-routing/pull/108.

We'll fix the "Search parameter names must be unique" kind of errors next.

rsmayda:
Released as part of Router-5.4.4: https://github.com/awslabs/fhir-works-on-aws-routing/pull/110/files

PR #108, opened 2021-07-30
fix: fix invalid capability statement

carvantes:
Fixes part of https://github.com/awslabs/fhir-works-on-aws-routing/issues/106

Note that `export-poll-status` was never a real operation per the FHIR spec. The `export` operation does both things(initiate export requests, and get the status/results of previous export jobs)


rsmayda:
Looks good I think we need to update unit tests

carvantes:
> Looks good I think we need to update unit tests

@rsmayda, My bad. I just updated them

rsmayda (approved):

PR #121, opened 2021-08-25
fix: add routing from server errors to conflict errors

ssvegaraju:
- fix: add routing from server errors to conflict errors
- update interface package

Issue #, if available:

Description of changes:


carvantes (changes requested):
Code changes look good, but there's an issue with the lock file.

`package-lock.json` shoulnd't be here. You should be updating `yarn.lock` instead. Remember to always use `yarn` instead of `npm` when working on FWoA

package.json
```diff
@@ -37,7 +37,7 @@
     "cors": "^2.8.5",
     "errorhandler": "^1.5.1",
     "express": "^4.17.1",
-    "fhir-works-on-aws-interface": "^11.0.0",
+    "fhir-works-on-aws-interface": "^11.1.0",
```
carvantes, line 40: You need to commit the corresponding  updates to yarn.lock

carvantes (changes requested):

package.json
ssvegaraju: Thanks for catching that, just fixed the `yarn.lock` in the latest commit.

carvantes (approved):

PR #131, opened 2021-09-17
fix: also sort by date for $docref when requesting latest doc

ghost:
Bug #132
Description of changes:
Added date to sort for $docref when getting latest.

### DIFF
--- a/src/operationDefinitions/USCoreDocRef/convertDocRefParamsToSearchParams.ts
+++ b/src/operationDefinitions/USCoreDocRef/convertDocRefParamsToSearchParams.ts
@@ -20,7 +20,7 @@ export const convertDocRefParamsToSearchParams = (docRefParams: DocRefParams) =>
 
     if (!start && !end) {
         // only latest document
-        searchParams = { _sort: '-date', _count: '1', ...searchParams };
+        searchParams = { _sort: '-period,-date', _count: '1', ...searchParams };
     } else {
         const dateParams = [];
         if (start) {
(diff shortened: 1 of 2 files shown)


carvantes:
Hi,
 
Our interpretation is that “period” should be used. Although the FHIR spec is actually a bit vague about this.
 
https://www.hl7.org/fhir/documentreference.html
From DocumentReference search params:
- `date`: _When this document reference was created._ This seems to be metadata (i.e. creation date) about the document
- `period`: _Time of service that is being documented._ This seems to refer to real world dates when treatment/service was provided
 
http://www.hl7.org/fhir/us/core/2018Jan/OperationDefinition-docref.html
 
For US Core $docref params:
`start`: _The date range relates to care dates, not record currency dates - e.g. all records relating to care provided in a certain date range. If no start date is provided, all documents prior to the end date are in scope. If neither a start date nor an end date is provided, the most recent or current document is in scope._
 
 
“the most recent or current document is in scope” is a bit ambiguous. It’s fairly clear that the start/end $docref params operate on “period”(since they specifically target “care dates, not record currency dates”), so we assume that “the most recent or current document” in this context refers to the document that contains the most recent care date, not the most recently created DocumentReference.
 
We believe that “period” should be used here. 
Still, please let us know if you find additional documentation that more accurately describes the behavior of $docref or if you have observed different behavior in other implementations. We’ll gladly make the adjustments to make our implementation correct.

carvantes (approved):

PR #138, opened 2021-10-12
feat: use requiresAccessToken value from GetExportStatusResponse

carvantes:
We are adding flexibility to how the export results URLs are generated and we need requiresAccessToken to be dynamic.
Currently the requiresAccessToken field is hard-coded to false in the router response.

Related PRs:
- https://github.com/awslabs/fhir-works-on-aws-interface/pull/85

### DIFF
--- a/src/router/routes/exportRoute.ts
+++ b/src/router/routes/exportRoute.ts
@@ -103,7 +103,7 @@ export default class ExportRoute {
                             queryParams,
                             groupId,
                         ),
-                        requiresAccessToken: false,
+                        requiresAccessToken: response.requiresAccessToken,
                         output: response.exportedFileUrls,
                         error: response.errorArray,
                     };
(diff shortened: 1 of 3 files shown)


ssvegaraju (dismissed):

src/router/routes/exportRoute.ts
```diff
@@ -103,7 +103,7 @@ export default class ExportRoute {
                             queryParams,
                             groupId,
                         ),
-                        requiresAccessToken: false,
+                        requiresAccessToken: response.requiresAccessToken,
```
rsmayda, line 106: I am surprised this compiles without the new interface

rsmayda (dismissed):

src/router/routes/exportRoute.ts
carvantes: +1, I also don't fully understand why this happens

rsmayda (approved):

PR #145, opened 2021-11-26
perf: Compile json validation schema at init instead of in the first call

organom:
When using the default JsonSchemaValidator to validate incoming POST and PUT calls, it was noticed that the first call would take longer than the subsequent ones (even with lambda provisioned concurrency).

This happens because the AJV library compiles the schema on the first validate if not compiled before.

**Description of changes:**
`COMPILE_SCHEMA` environment variable is added that when set to a value of `true`, compiles the Schema on the constructor.

Running the tests can easily prove the behavior:

- COMPILE_SCHEMA not set or different than `true`

Screenshot 2021-11-26 at 15 20 41

- COMPILE_SCHEMA set to `true`

Screenshot 2021-11-26 at 15 20 26

### DIFF
--- a/src/router/validation/jsonSchemaValidator.ts
+++ b/src/router/validation/jsonSchemaValidator.ts
@@ -23,11 +23,11 @@ export default class JsonSchemaValidator implements Validator {
         let schema;
         if (fhirVersion === '4.0.1') {
             ajv.addMetaSchema(schemaDraft06);
-            ajv.addSchema(fhirV4Schema);
+            ajv.compile(fhirV4Schema);
             schema = fhirV4Schema;
         } else if (fhirVersion === '3.0.1') {
             ajv.addMetaSchema(schemaDraft04);
-            ajv.addSchema(fhirV3Schema);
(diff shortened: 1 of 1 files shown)


Bingjiling:
Hi @organom , 

Thanks for diving deep into this cold start behavior and contributing back to FWoA! 

I like the approach taken here. I don't think the environment variable `COMPILE_SCHEMA` is needed though, as it is improving performance without bringing side effect. Another nit-pick, could you please use `perf:` prefix for the commit message? (We have an automatic workflow to create a release pr if `feat:` prefix is used)

Thanks,
Yanyu

Bingjiling (changes requested):
Please remove environment variable `COMPILE_SCHEMA` and use `perf:` for commit message prefix.

organom:
@Bingjiling Thank you for the input and sorry for the delay on making the changes. They are now done as requested :)

Bingjiling (approved):
LGTM! Thanks again for your contribution 👍  Could you please update the branch to be up to date with mainline?

Issue #146, opened 2021-12-06
[Feature Request] MIME type application/fhir+json


AllanHodkinson:
Hi Robert & team,
One of our team members had a query today about the MIME type returned by the routing module, in that it is `application/json` rather than `application/fhir+json` as the specification suggests _should_ be used (https://www.hl7.org/fhir/http.html#mime-type).

This is clearly not a show stopper as it's never been raised as an issue on here, or by any of our customers, but I'm wondering if you'd had any thoughts about this?

Many thanks,
Allan

Bingjiling:
Hi @AllanHodkinson , 

Thanks for reaching out! We'll look into updating the MIME type. 

Thanks,
Yanyu

AllanHodkinson:
Thanks Yanyu, there is some guidance on response behaviour for different scenarios on that web page I've linked above.

Bingjiling:
Hi Allan,

This issue has been fixed and released in [v6.3.1](https://github.com/awslabs/fhir-works-on-aws-routing/releases/tag/v6.3.1).   Thanks for you help identifying this issue! 

Cheers,
Yanyu

PR #147, opened 2021-12-08
fix: Use application/fhir+json as default content-type

Bingjiling:
Issue #, if available:
#146 

Description of changes:
* Use `application/fhir+json` as default content-type
* Use `application/json` if request content-type is `application/json`
* Deployed and tested in AWS account 

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -83,6 +83,18 @@ export function generateServerlessRouter(
 
     mainRouter.use(setServerUrlMiddleware(fhirConfig));
 
+    // Set default content-type to 'application/fhir+json'
+    mainRouter.use(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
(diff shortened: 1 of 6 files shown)


src/app.ts
```diff
@@ -83,6 +83,18 @@ export function generateServerlessRouter(
 
     mainRouter.use(setServerUrlMiddleware(fhirConfig));
 
+    // Set default content-type to 'application/fhir+json'
+    mainRouter.use(async (req: express.Request, res: express.Response, next: express.NextFunction) => {
```
carvantes: Can you move this middleware to it's own file under https://github.com/awslabs/fhir-works-on-aws-routing/tree/mainline/src/router/middlewares?
It feels a bit better organized that way.

carvantes (approved):

PR #149, opened 2021-12-14
chore: add unit test coverage and enforcement

ssvegaraju:
Issue #, if available:

Description of changes:

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -1,0 +1,17 @@
+#
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  SPDX-License-Identifier: Apache-2.0
+#
+
+name: Code Coverage Report
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2 # Check out your repository
+    - run: yarn install # Install dependencies
+    - run: yarn test --coverage # Run test
+    - run: bash <(curl -s https://codecov.io/bash) # Upload to Codecov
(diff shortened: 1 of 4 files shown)


codecov.yml
```diff
@@ -0,0 +1,19 @@
+coverage:
+  round: down
+  precision: 2
+  range: '90...100'
+  status:
+    project:
+      default:
+        # basic
+        target: 90%
```
carvantes: It looks like the codecov setup is not working.

These same changes were done in the search package (https://github.com/awslabs/fhir-works-on-aws-search-es/pull/147) but there are no Codecov reports yet: https://app.codecov.io/gh/awslabs/fhir-works-on-aws-search-es

Also, Codecov is not posting coverage messages on PRs like https://github.com/awslabs/service-workbench-on-aws/pull/844#issuecomment-998888816

---

Looks like we cannot use  target of 90% for all our packages. Looking at https://github.com/awslabs/fhir-works-on-aws-routing/runs/4529235534?check_suite_focus=true, this package has  test coverage of 75%. Also, looks like the coverage target is not being enforced (I think this PR should fail the 90% goal). We may need to set up different coverage goals for each package to get started.

.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,17 @@
...
+#  Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#  SPDX-License-Identifier: Apache-2.0
+#
+
+name: Code Coverage Report
+
+on: [push, pull_request]
+
+jobs:
+  test:
```
carvantes: consider a more meaningful job name. Maybe `Report test coverage`

carvantes (changes requested):

.github/workflows/code-coverage.yml
ssvegaraju: Updated the name!

codecov.yml
ssvegaraju: Looks like the issue was the coverageReporters defined in the package.json. Removing that and updating coverage seems to have worked.

carvantes (approved):

Issue #136, opened 2021-10-01
[Feature Request] Log cloudfront request headers for downstream (Goelocation) analytics 


medhost-jshaver:
**Is your feature request related to a problem? Please describe.**

We have a need to perform downstream analytics on the requests made to fhirworks and as part of this analytics we need data around geolocation. API Gateway appears to have cloudfront headers available as part of the request and these cloudfront headers contain all the data we require.

Cloudfront request headers that we are interested in:
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-cloudfront-headers.html

**Describe the solution you'd like**
- Log request headers as part of routing package
- Selectively log header values that are not sensitive in nature
- Perhaps make this conditional logging based on an environment variable?

**Describe alternatives you've considered**
- Using API gateway execution logs, but the 1024 byte limit truncated the request information
This also logs the sensitive header attributes like JWT token
- Using API gateway access logs, but these only appear to have access to the "context" object associated with the request coming to API gateway. The cloudfront headers do not appear to be accessible from this context object. See here: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-logging-variables.html

nguyen102:
Thank you for creating this feature request, we've added this to our team's backlog.

kcadette:
FHIR Works on AWS has been moved to maintenance mode. While in maintenance, we will not add any new features to this solution. All security issues should be reported directly to AWS Security at [aws-security@amazon.com](mailto:security@amazon.com). If you are new to this solution, we advise you to explore using [HealthLake](https://aws.amazon.com/healthlake), which is our managed service for building FHIR based transactional and analytics applications. You can get started by contacting your AWS Account team. If you are an existing customer
of FHIR Works on AWS, and have additional questions or need immediate help, please reach out to [fwoa-migration-support@amazon.com](mailto:fwoa-migration-support@amazon.com) or contact your AWS Account team.

PR #151, opened 2021-12-23
feat: log cloudfront request headers

ssvegaraju:
Issue #136, if available:

Description of changes:
Modified fhir-service lambda in deployment package to also pass headers to routing package (from `event.headers`). This gives access to the CloudFront request headers that are attached by APIGateway. We may need to whitelist additional geolocation headers as well, as only `CloudFront-Viewer-Country` is present at the moment, and I'm not sure if that is sufficient for the ask.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -28,7 +28,7 @@ import { initializeOperationRegistry } from './operationDefinitions';
 import { setServerUrlMiddleware } from './router/middlewares/setServerUrl';
 import { setTenantIdMiddleware } from './router/middlewares/setTenantId';
 import { setContentTypeMiddleware } from './router/middlewares/setContentType';
-import getComponentLogger from './loggerBuilder'
+import getComponentLogger from './loggerBuilder';
 
 const configVersionSupported: ConfigVersion = 1;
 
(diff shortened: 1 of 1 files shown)


carvantes:
> Modified fhir-service lambda in deployment package to also pass headers to routing package (from event.headers). 

Does this PR depends on changes to the deployment package?

src/app.ts
```diff
@@ -51,6 +52,7 @@ function prepareRequestContext(req: express.Request): RequestContext {
 export function generateServerlessRouter(
     fhirConfig: FhirConfig,
     supportedGenericResources: string[],
+    headers?: any,
```
carvantes, line 55: Avoid using `any`. maybe use a more specific param name.

Also, use an optional options object as the last param so that it's easier to add more optional params in the future. 
Similar to: 
https://github.com/awslabs/fhir-works-on-aws-search-es/blob/30290fdbe15fac047b0b0a76bd4fea7e60f3d62a/src/elasticSearchService.ts#L84-L96

carvantes (changes requested):

PR #152, opened 2022-01-14
feat: add subscription validator

Bingjiling:
Issue #, if available:

Description of changes:

Add subscription validator

### DIFF
--- a/src/router/validation/subscriptionValidator.test.ts
+++ b/src/router/validation/subscriptionValidator.test.ts
@@ -5,7 +5,7 @@
 
 import { InvalidResourceError, Search } from 'fhir-works-on-aws-interface';
 import ElasticSearchService from '../__mocks__/elasticSearchService';
-import SubscriptionValidator, { GetAllowEndpoints, SubscriptionEndpoint } from './subscriptionValidator';
+import SubscriptionValidator, { SubscriptionEndpoint } from './subscriptionValidator';
 import validPatient from '../../../sampleData/validV4Patient.json';
 import invalidPatient from '../../../sampleData/invalidV4Patient.json';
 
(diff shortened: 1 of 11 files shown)


src/router/validation/subscriptionValidator.ts
```diff
@@ -0,0 +1,135 @@
...
+        } else {
+            if (tenantId === undefined) {
+                return this.allowListMap[SINGLE_TENANT_ALLOW_LIST_KEY];
+            }
+            throw new Error('This instance has multi-tenancy disabled, but the incoming request has a tenantId');
+        }
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    private filterResourcesToValidate(resource: any): any[] {
```
carvantes: nit: rename to something like `extractSubscriptionResources`

src/router/validation/subscriptionValidator.ts
```diff
@@ -0,0 +1,135 @@
...
+        const { resourceType } = resource;
+        let resourcesToValidate: any[] = [];
+        if (resourceType === SUBSCRIPTION_RESOURCE_TYPE) {
+            resourcesToValidate = [resource];
```
carvantes: you can simply `return [resource]` here. In general it looks to me that `resourcesToValidate` is redundant and can be removed

src/router/validation/subscriptionValidator.ts
```diff
@@ -0,0 +1,135 @@
...
+                throw new InvalidResourceError(
+                    `Subscription resource is not valid. Endpoint ${res.channel.endpoint} is not allow listed.`,
+                );
+            }
+            this.search.validateSubscriptionSearchCriteria(res.criteria);
+        });
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    private isEndpointAllowListed(allowList: (string | RegExp)[], endpoint: string): boolean {
```
carvantes: Let's not disable `class-methods-use-this`. this method can be changed to a fuction outisde of the class. ([example](https://github.com/awslabs/fhir-works-on-aws-search-es/blob/9847e1415d5067496b391e733fdd365185a062f0/src/FHIRSearchParametersRegistry/index.ts#L67))

I know that our codebase disables `class-methods-use-this` a lot, but going forward we should avoid it.

src/router/validation/subscriptionValidator.ts
```diff
@@ -0,0 +1,135 @@
...
+            throw new Error('This instance has multi-tenancy enabled, but the incoming request is missing tenantId');
+        } else {
+            if (tenantId === undefined) {
+                return this.allowListMap[SINGLE_TENANT_ALLOW_LIST_KEY];
+            }
+            throw new Error('This instance has multi-tenancy disabled, but the incoming request has a tenantId');
+        }
+    }
+
+    // eslint-disable-next-line class-methods-use-this
```
carvantes: Let's not disable `class-methods-use-this.`

src/router/validation/subscriptionValidator.test.ts
```diff
@@ -0,0 +1,208 @@
...
+                request: {
+                    method: 'DELETE',
+                    url: 'Patient/some-fake-id',
+                },
+            },
+        ],
+    };
+};
+
+describe('Multi-tenancy mold: Validating Subscriptions', () => {
```
carvantes: typo? `mold`

src/router/validation/subscriptionValidator.test.ts
```diff
@@ -0,0 +1,208 @@
...
+        const subscription = getSubscriptionResource('https://fake-end-point-tenant1');
+        await expect(subscriptionValidator.validate(subscription)).rejects.toThrowError(
+            new InvalidResourceError(
+                'This instance has multi-tenancy enabled, but the incoming request is missing tenantId',
+            ),
+        );
+    });
+});
+
+describe('Single-tenancy mold: Validating Subscriptions', () => {
```
carvantes: typo? `mold`

src/router/validation/subscriptionValidator.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    test('No error when validating resources that are not Subscription or Bundle', async () => {
+        await expect(subscriptionValidator.validate(invalidPatient, 'tenant1')).resolves.toEqual(undefined);
+    });
+
+    test('No error when validating Bundle that does not contain Subscription resource', async () => {
+        const bundle = getBundleResource(validPatient);
+        await expect(subscriptionValidator.validate(bundle, 'tenant2')).resolves.toEqual(undefined);
+    });
+
+    test('Show error when validating invalid Subscription resource', async () => {
```
carvantes: Can you add  couple more tests like this?
e.g. channel.type must be rest-hook, channel.payload must be application/fhir+json if present. etc.

src/router/validation/subscriptionValidator.test.ts
carvantes: I think that there's a bit too much repetition of tests for multi and single tenancy. We should find a way to only write them once. 

Try using `test.each` or `describe.each` (https://jestjs.io/docs/api#testeachtablename-fn-timeout) There are some examples already in our codebase.

src/router/validation/subscriptionValidator.ts
```diff
@@ -0,0 +1,135 @@
...
+export default class SubscriptionValidator implements Validator {
+    private ajv: Ajv.Ajv;
+
+    private validateJSON: Ajv.ValidateFunction;
+
+    private search: Search;
+
+    private allowListMap: { [key: string]: (string | RegExp)[] } = {};
+
+    private allowListLoaded = false;
```
carvantes: I think it would be simpler to just have `SubscriptionValidator` receive `SubscriptionEndpoint[]` as param instead of having to call a fn and caching the result.

That way the client (most likely the `config.ts` file in fwoa-deployment) can fetch the `SubscriptionEndpoint` array and construct the Validator. In the default setup the allow list can be fetched just once at server startup(when building FhirConfig).

Maybe I'm missing something, is there a reason to have the fn call in this class?

carvantes (changes requested):

src/router/validation/subscriptionValidator.ts
Bingjiling: Yes, it would be simpler to just pass in the endpoints. I didn't do it as there's no `await` used in `config.ts`, and I thought it was done on purpose to avoid top-level await problems. Maybe I'm overthinking this?

carvantes: `await` won't be a problem. A few months ago we made `getFhirConfig` async so that customers can build their config asynchronously. The `index.ts` already does `await getFhirConfig()` (even if by default all config is static)

https://github.com/awslabs/fhir-works-on-aws-deployment/pull/464

src/router/validation/subscriptionValidator.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    private readonly validateJSON: Ajv.ValidateFunction;
+
+    private search: Search;
+
+    private allowListMap: { [key: string]: (string | RegExp)[] } = {};
+
+    private readonly enableMultiTenancy: boolean;
+
+    constructor(search: Search, allowList: SubscriptionEndpoint[], enableMultiTenancy: boolean) {
```
carvantes: Can you make the last param an optional object and have `enableMultiTenancy` default to false?
e.g. https://github.com/awslabs/fhir-works-on-aws-persistence-ddb/blob/e2ccf527eeb5fede5ac0a1b42511ec4ab08b7255/src/objectStorageService/s3DataService.ts#L42

The different classes that use `enableMultiTenancy` follow that pattern. Plus, it's easier to understand ( it's hard to tell the meaning of true in `new SubscriptionValidator(search, multiTenantAllowList, true);`)

src/router/validation/subscriptionValidator.test.ts
```diff
@@ -0,0 +1,202 @@
...
+                ),
+            );
+        });
+
+        test('Show error when payload is not application/fhir+json', async () => {
+            const subscription = getSubscriptionResource(endpoint);
+            subscription.channel.payload = 'application/xml';
+            await expect(validator.validate(subscription, tenantId)).rejects.toThrowError(
+                new InvalidResourceError(
+                    'Subscription resource is not valid. Error was: data.channel.payload should be equal to constant',
```
carvantes: you can use `ajv-errors` and the `errorMessage` field to have better error messages.

e.g. https://github.com/awslabs/fhir-works-on-aws-routing/blob/836719f7bdc4f424ad7173695cf8a8b96dcb7d3f/src/operationDefinitions/USCoreDocRef/parseParams.ts#L83

the ajv default error of "should be equal to constant" is not great for customers

carvantes (commented):
Looks good

I just left a couple comments focused on usability.

carvantes (approved):

PR #157, opened 2022-02-24
feat: allow configuring the max number of active subscriptions

carvantes:
300 is an ok default but it may not be enough for some customers. It's better to let them configure it


ssvegaraju (approved):
LGTM

PR #162, opened 2022-03-25
feat: add Batch support

ssvegaraju:
Issue #, if available:

Description of changes:
Added logic to parse requests made to the root URL with a Batch request. Contains logic to validate the bundle according to https://www.hl7.org/fhir/http.html#brules, and adds unit tests for batch. This PR is in draft until the persistence changes are released and updated on the interface package.

We use a limit of 750 requests per bundle for the following reason:
Lambda has a payload limit of 6MB. Taking an average request size of 4KB, this gives us that 6MB/4KB = 1500 requests. To give some leeway, we divide by half to get 750.

TESTING:
tested through deployment and querying for various different bundle configurations such as:
- batches with references to resource within the bundle
- all different CRUD operations
- failing CRUD operations should not affect bundle success

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -4,3 +4,6 @@
  */
 
 export const MAX_BUNDLE_ENTRIES = 25;
+// Lambda payload limit is 6MB, assuming an average request of 4KB,
+// we have 6MB / 4Kb = 1500. Dividing by half to allow for contingencies, we have the following limit:
+export const MAX_BATCH_ENTRIES = 750;
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -4,8 +4,3 @@
  */
 
 export const MAX_BUNDLE_ENTRIES = 25;
-// Lambda payload limit is 6MB, assuming an average request of 4KB,
-// we have 6MB / 4Kb = 1500. Dividing by half to allow for contingencies, we get 750.
-// This value is customizable during deployment by appending --maxBatchEntries <your value here>
-// if no param is defined during deployment, we use this default
-export const MAX_BATCH_ENTRIES = Number(process.env.MAX_BATCH_ENTRIES) || 750;
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -5,5 +5,7 @@
 
 export const MAX_BUNDLE_ENTRIES = 25;
 // Lambda payload limit is 6MB, assuming an average request of 4KB,
-// we have 6MB / 4Kb = 1500. Dividing by half to allow for contingencies, we have the following limit:
-export const MAX_BATCH_ENTRIES = 750;
+// we have 6MB / 4Kb = 1500. Dividing by half to allow for contingencies, we get 750.
+// This value is customizable during deployment by appending --maxBatchEntries <your value here>
+// if no param is defined during deployment, we use this default
+export const MAX_BATCH_ENTRIES = Number(process.env.MAX_BATCH_ENTRIES) || 750;
--- a/src/router/bundle/bundleGenerator.ts
+++ b/src/router/bundle/bundleGenerator.ts
@@ -129,4 +129,49 @@ export default class BundleGenerator {
         response.entry = entries;
         return response;
     }
+
+    static generateBatchBundle(baseUrl: string, bundleEntryResponses: BatchReadWriteResponse[]) {
+        const id = uuidv4();
+        const response = {
+            resourceType: 'Bundle',
+            id,
+            type: 'batch-response',
+            link: [
+                {
+                    relation: 'self',
+                    url: baseUrl,
+                },
+            ],
+            entry: [],
+        };
+
+        const entries: any = [];
+        bundleEntryResponses.forEach((bundleEntryResponse) => {
+            let status = '200 OK';
+            if (bundleEntryResponse.operation === 'create') {
+                status = '201 Created';
+            } else if (
+                ['read', 'vread'].includes(bundleEntryResponse.operation) &&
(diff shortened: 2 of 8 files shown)


src/router/bundle/bundleGenerator.ts
```diff
@@ -129,4 +129,51 @@ export default class BundleGenerator {
...
+            entry: [],
+        };
+
+        const entries: any = [];
+        bundleEntryResponses.forEach((bundleEntryResponse) => {
+            let status = '200 OK';
+            if (bundleEntryResponse.operation === 'create') {
+                status = '201 Created';
+            } else if (
+                ['read', 'vread'].includes(bundleEntryResponse.operation) &&
```
Bingjiling: I thought vread is not supported? Also why would `isEmpty(bundleEntryResponse.resource)` map to 403 instead of 404?

src/router/bundle/bundleGenerator.ts
```diff
@@ -129,4 +129,51 @@ export default class BundleGenerator {
...
+                isEmpty(bundleEntryResponse.resource)
+            ) {
+                status = '403 Forbidden';
+            } else if (bundleEntryResponse.operation === undefined || bundleEntryResponse.vid === '') {
+                status = '404 Not Found';
```
Bingjiling: Using vid to indicate if there's error thrown in read is not a good practice. Also if error is thrown in read, it does not always mean the record is not found, there could be error of other kind as well. Add logic in Persistence to differentiate 404 from other errors?

src/router/bundle/bundleGenerator.ts
```diff
@@ -129,4 +129,51 @@ export default class BundleGenerator {
         response.entry = entries;
         return response;
     }
+
+    static generateBatchBundle(baseUrl: string, bundleEntryResponses: BatchReadWriteResponse[]) {
```
Bingjiling, line 133: Add unit test for different status code?

src/router/bundle/bundleHandler.ts
```diff
@@ -58,17 +58,100 @@ export default class BundleHandler implements BundleHandlerInterface {
...
         tenantId?: string,
     ) {
-        throw new createError.BadRequest('Currently this server only support transaction Bundles');
+        const startTime = new Date();
+
+        await validateResource(this.validators, bundleRequestJson, { tenantId });
+
+        let requests: BatchReadWriteRequest[];
+        try {
+            // TODO use the correct persistence layer
```
Bingjiling: What does this TODO mean?

src/router/bundle/bundleHandler.ts
```diff
@@ -58,17 +58,100 @@ export default class BundleHandler implements BundleHandlerInterface {
         this.validators = validators;
     }
 
-    /* eslint-disable  @typescript-eslint/no-unused-vars,prettier/prettier */
     async processBatch(
```
Bingjiling, line 61: Most of the logic in this method looks similar as processTransaction, try re-using existing logic and avoid writing the same thing twice.

ssvegaraju: I refactored all similar code out into separate methods, and added a constant to limit the size of batch bundles. I also added more tests for the batch bundle in error cases. I believe we use 403 Forbidden to indicate a resource that was returned but not part of the bundle because it was filtered out.

src/constants.ts
```diff
@@ -4,3 +4,6 @@
  */
 
 export const MAX_BUNDLE_ENTRIES = 25;
+// Lambda payload limit is 6MB, assuming an average request of 4KB,
+// we have 6MB / 4Kb = 1500. Dividing by half to allow for contingencies, we have the following limit:
+export const MAX_BATCH_ENTRIES = 750;
```
Bingjiling: Can we make this a value that can be passed in from deployment package with a default of 750? Customers might want to customize this number.

ssvegaraju: Done! I added it to the changes in the deployment package as well.

src/router/bundle/bundleGenerator.ts
Bingjiling: This method is very similar to `generateTransactionBundle`, combine common codes?

src/router/bundle/bundleGenerator.ts
```diff
@@ -129,4 +129,56 @@ export default class BundleGenerator {
...
+            const entry: any = {
+                response: {
+                    status,
+                    location: `${bundleEntryResponse.resourceType}/${bundleEntryResponse.id}`,
+                    etag: bundleEntryResponse.vid,
+                    lastModified: bundleEntryResponse.lastModified,
+                },
+            };
+            // remove unnecessary fields from response
+            if (bundleEntryResponse.error) {
```
Bingjiling: Put this logic in Persistence instead of here? If etag and lastModified should not exist, it is Persistence's responsibility to remove them instead of depending on another package to handle it.

src/router/bundle/bundleHandler.ts
Bingjiling: Looks good!

Bingjiling (commented):
Looks good! Some last changes for `generateBatchBundle` and we're good to go.

src/router/bundle/bundleGenerator.ts
ssvegaraju: I don't think it's possible to remove the vid/lastmodified fields from the persistence response as they are a required property of the type. However, I removed the logic to explicitly remove them, and instead combined the logic with the transaction bundle generation.

Bingjiling (approved):

Issue #167, opened 2022-05-23
change content-type to accept header


liquid36:
Hi!
I think that is better to check for de **Accept** header rather than the **ContentType** header.  Accept header is about what the client expect. 

https://github.com/awslabs/fhir-works-on-aws-routing/blob/cb044c37620248a0bd5f261761dd9dc09dbfa426/src/router/middlewares/setContentType.ts#L16

Particulary, i'm going throw ONC Certification, and the **$export test** expect to have an _application/json_ response, not a _application/fhir+josn_ response. 

What do you think?

Bingjiling:
Hi @liquid36,

Thanks for spotting the issue! I've checked the FHIR spec and you're right about this one. As this is a one-line change, do you want to take the opportunity and create a PR for it? 

Thanks,
Yanyu

liquid36:
Yes, no problem. 

#168

PR #168, opened 2022-05-23
fix: change content-type to accept header

liquid36:
Issue #167 

Description of changes:

Update `setContentTypeMiddleware` in order to user the **Accept** header.

The "Accept" header field can be used by user agents to specify response media types that are acceptable.

### DIFF
--- a/src/router/middlewares/setContentType.test.ts
+++ b/src/router/middlewares/setContentType.test.ts
@@ -33,7 +33,7 @@ describe('setContentTypeMiddleware', () => {
         const nextMock = jest.fn();
         const req = {
             headers: {
-                'content-type': 'application/json',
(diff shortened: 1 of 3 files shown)


src/router/middlewares/setContentType.test.ts
```diff
@@ -33,7 +33,7 @@ describe('setContentTypeMiddleware', () => {
         const nextMock = jest.fn();
```
Bingjiling, line 33: Update test description to 'request should return application/json if user request application/json in accept header'

Bingjiling (commented):
Looks good! Could you please update the failed unit tests as well?

Bingjiling (approved):

PR #172, opened 2022-09-08
fix: Throw error if url and POSTed resourceType do not match

rsmayda:
Issue #, if available: https://github.com/awslabs/fhir-works-on-aws-deployment/issues/682

Description of changes: validating that url resourceType and resource's resourceType is the same


ssvegaraju (approved):
LGTM!

PR #192, opened 2023-01-13
[WIP] Add Fixes to Routing Package

kcadette:
Issue #, if available:

Description of changes:
Improve Error Handling. Including Bad Request and Method Not Allowed errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "cors": "^2.8.5",
     "errorhandler": "^1.5.1",
     "express": "^4.17.1",
-    "fhir-works-on-aws-interface": "^12.1.0",
+    "fhir-works-on-aws-interface": "^12.1.1",
     "flat": "^5.0.0",
     "http-errors": "^1.8.0",
     "lodash": "^4.17.15",
--- a/src/router/routes/errorHandling.ts
+++ b/src/router/routes/errorHandling.ts
@@ -11,7 +11,7 @@ import {
     isInvalidSearchParameterError,
     isResourceConflictError,
     isBadRequestError,
-    isMethodNotAllowed
+    isMethodNotAllowed,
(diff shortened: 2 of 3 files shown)


ssvegaraju:
I think you may need to run `yarn install` after bumping the package version to get the checks to pass

kcadette:
unfortunately, the latest version isn't an option when I run yarn install. Its also not updated on npm yet so there could be some lag there

src/router/routes/errorHandling.ts
```diff
@@ -28,7 +34,7 @@ export const applicationErrorMapper = (
         next(new createError.NotFound(err.message));
         return;
     }
-    if (isInvalidResourceError(err)) {
+    if (isInvalidResourceError(err) || isBadRequestError(err)) {
```
Bingjiling, line 37: If isInvalidResourceError is treated the same way as isBadRequestError from the customers' perspective, what's the gain of differentiating the two?

package.json
```diff
@@ -38,7 +38,7 @@
     "cors": "^2.8.5",
     "errorhandler": "^1.5.1",
     "express": "^4.17.1",
-    "fhir-works-on-aws-interface": "^12.1.0",
+    "fhir-works-on-aws-interface": "^12.1.1",
```
Bingjiling, line 41: Please check in the corresponding change in the lock file as well.

src/router/routes/errorHandling.ts
sevu1: It would be a bit strange to use isInvalidResourceError for all bad request errors, I think BadRequestError and we can deprecate isInvalidResourceError later

sevu1 (approved):
It would be nice to add description to PR, otherwise - looks good

PR #193, opened 2023-01-23
fix: add optional input validation

ssvegaraju:
Issue #, if available:

Description of changes:
Added optional input validation configuration. Tested by running integration tests and adding new integ tests.

### DIFF
--- a/src/router/handlers/resourceHandler.test.ts
+++ b/src/router/handlers/resourceHandler.test.ts
@@ -1072,7 +1072,7 @@ describe('Testing xhtml validation', () => {
         const validatedPatient = validateXHTMLResource(validPatient);
 
         // CHECK
-        expect(validatedPatient).toMatchObject(validPatient);
+        expect(validatedPatient).toBe(true);
     });
 
     test('invalid patient resource is filtered', () => {
@@ -1094,6 +1094,6 @@ describe('Testing xhtml validation', () => {
         const validatedPatient = validateXHTMLResource(scriptedPatient);
 
         // CHECK
-        expect(validatedPatient).toMatchObject(validPatient);
+        expect(validatedPatient).toBe(false);
     });
 });
--- a/src/router/handlers/resourceHandler.test.ts
+++ b/src/router/handlers/resourceHandler.test.ts
@@ -1065,9 +1065,6 @@ describe('Testing history', () => {
 
 describe('Testing xhtml validation', () => {
     test('valid patient resource is not affected', () => {
-        if (process.env.VALIDATE_XHTML !== 'true') {
-            return;
-        }
         // BUILD & OPERATE
         const validatedPatient = validateXHTMLResource(validPatient);
 
@@ -1076,9 +1073,6 @@ describe('Testing xhtml validation', () => {
     });
 
     test('invalid patient resource is filtered', () => {
-        if (process.env.VALIDATE_XHTML !== 'true') {
-            return;
-        }
         // BUILD
         const scriptedPatient = {
             ...validPatient,
--- a/src/router/handlers/resourceHandler.test.ts
+++ b/src/router/handlers/resourceHandler.test.ts
@@ -34,6 +34,7 @@ import OperationsGenerator from '../operationsGenerator';
 import ElasticSearchService from '../__mocks__/elasticSearchService';
 import DynamoDbDataService from '../__mocks__/dynamoDbDataService';
 import JsonSchemaValidator from '../validation/jsonSchemaValidator';
+import { validateXHTMLResource } from './utils';
 
 const enum SEARCH_PAGINATION_PARAMS {
     PAGES_OFFSET = '_getpagesoffset',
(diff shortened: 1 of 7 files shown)


src/router/handlers/utils.ts
```diff
@@ -5,5 +5,31 @@
...
+            return `${name}=${value}`;
+        }
+    },
+};
+const xssValidator = new xss.FilterXSS(options);
+
+export const hash = (o: any): any => createHash('sha256').update(JSON.stringify(o)).digest('hex');
+
+export const validateXHTMLResource = (resource: any): any => {
+    if (process.env.VALIDATE_XHTML === 'true') {
```
Bingjiling: Does the validator strip off things that does not pass validate or does it fail request that does not pass validate? If we're enforcing HTML validate, we should just fail request that contains HTML. If we save the stripped version, it'll give requester the false impression that the request went through successfully and all contents were saved.

src/router/handlers/resourceHandler.ts
```diff
@@ -45,16 +45,27 @@ export default class ResourceHandler implements CrudHandlerInterface {
     }
 
     async create(resourceType: string, resource: any, tenantId?: string) {
-        await validateResource(this.validators, resourceType, resource, { tenantId, typeOperation: 'create' });
+        const strippedResource = validateXHTMLResource(resource);
+        await validateResource(this.validators, resourceType, strippedResource, { tenantId, typeOperation: 'create' });
```
Bingjiling: Can `validateXHTMLResource` be called in `validateReousrce` instead? It's a more logical organization of methods since it is a validation functionality, it also avoids calling it three times from different methods.

src/router/handlers/utils.ts
```diff
@@ -5,5 +5,31 @@
  */
 
 import { createHash } from 'crypto';
+import * as xss from 'xss';
 
-export const hash = (o: any): string => createHash('sha256').update(JSON.stringify(o)).digest('hex');
+const options = {
+    whiteList: {
```
Bingjiling: Nit-pick: non-inclusive text. Although it's hard to change it since it's an attribute from a dependency.

src/router/handlers/utils.ts
```diff
@@ -5,5 +5,31 @@
  */
 
 import { createHash } from 'crypto';
+import * as xss from 'xss';
 
-export const hash = (o: any): string => createHash('sha256').update(JSON.stringify(o)).digest('hex');
+const options = {
```
Bingjiling: Can we check with security to see if this is the recommended library and setup to use? I remember an issue from another repo where a front end package was not sanitizing HTML as expected. Just want to make sure we're using the right package.

src/router/handlers/utils.ts
ssvegaraju: This package only strips or escapes invalid requests, I couldn't find a clean way to throw an Error instead when values were stripped without inefficient string comparisons. However: in the case where a `patient.name[0].family` is set to `'<script>alert(123);</script>'`, the request will fail because the validation will leave the field empty, which fails the validator check. However, `'<script>alert(123);</script>test'` would leave the field as `test` and be accepted.

src/router/handlers/resourceHandler.ts
ssvegaraju: I did originally have it there, the main issue I was running into was that it isn't possible to modify the resource in that method, which was the route I went because the package doesn't have a way I could find to throw errors instead. I will investigate further with security to see if there's a better choice out there.

Bingjiling: Got it, thanks for looking into it.

src/router/handlers/utils.ts
Bingjiling: Got it. Sounds like choosing another library that throws error is the most efficient way to go. Would comparing originalResource with validatedResource as strings work?

ssvegaraju: This ended up working with the caveat of removing the `text` field from resources before comparing them.

src/router/handlers/utils.ts
ssvegaraju: Switched to the more reputable sanitize-html package

src/router/handlers/resourceHandler.test.ts
```diff
@@ -1061,3 +1062,38 @@ describe('Testing history', () => {
         });
     });
 });
+
+describe('Testing xhtml validation', () => {
+    test('valid patient resource is not affected', () => {
+        if (process.env.VALIDATE_XHTML !== 'true') {
```
sevu1: I think for tests we should be validating resources whether or not `process.env.VALIDATE_XHTML === 'true'`

sevu1 (dismissed):

src/router/handlers/resourceHandler.test.ts
ssvegaraju: Good point, updated! I also realized that the utility function itself was also checking the env var in addition to the place it's actually used, so I updated that to be consistent with the tests as well.

sevu1 (approved):

PR #196, opened 2023-01-26
perf: improve regex performance with length limitation

Bingjiling:
Issue #, if available:

Description of changes:

* Add length limitation for regex to address CodeQL scan discovery 


src/router/bundle/bundleParser.test.ts
```diff
@@ -80,6 +80,71 @@ describe('parseResource', () => {
...
+                ],
+            };
+
+            try {
+                // OPERATE
+                await BundleParser.parseResource(bundleRequestJson, DynamoDbDataService, serverUrl);
+            } catch (e) {
+                // CHECK
+                expect((e as any).name).toEqual('Error');
+                expect((e as any).message).toEqual('Entry full URL length exceeds length limit of 500');
```
ssvegaraju: Maybe we shouldn't hardcode the limit value here, either we could use the CONST value directly or just say it exceeds the length limit here. I think the latter is a better approach as it gives less information to any malicious attackers. Same for the above test on reference URL

src/router/bundle/bundleParser.ts
```diff
@@ -334,6 +336,9 @@ export default class BundleParser {
 
         const references: Reference[] = referencePaths.map((referencePath) => {
             const entryReference = get(entry.resource, referencePath);
+            if (entryReference.length > MAX_REFERENCE_URL_LENGTH) {
+                throw new Error(`Reference URL exceeds length limit of ${MAX_REFERENCE_URL_LENGTH}`);
```
ssvegaraju: same nitpick as above with including the length limit value

ssvegaraju (commented):
LGTM! Just a minor comment:

ssvegaraju (dismissed):

src/regExpressions.test.ts
```diff
@@ -64,3 +70,41 @@ describe('captureFullUrlParts', () => {
...
+    test('Success match', () => {
+        const resourceString = `Patient/12345678`;
+        const actualMatch = resourceString.match(captureResourceTypeRegExp);
+
+        const expectedMatch = ['Patient/12345678', 'Patient'];
+
+        for (let i = 0; i < expectedMatch.length; i += 1) {
+            // @ts-ignore
+            expect(actualMatch[i]).toEqual(expectedMatch[i]);
+        }
```
nguyen102, line 84: Couldn't you just do `expect(actualMatch).toEqual(expectedMatch)`

src/router/bundle/bundleParser.ts
```diff
@@ -353,9 +358,14 @@ export default class BundleParser {
                 let rootUrl = fullUrlMatch[1];
                 // If the reference doesn't have a urlRoot, check if the entry's fullUrl has a urlRoot
                 if (rootUrl === undefined) {
-                    if (entry.fullUrl && entry.fullUrl.match(captureFullUrlParts)) {
+                    if (entry.fullUrl) {
```
nguyen102: Thoughts on avoiding the deeply nested if statement by breaking the code into two if statements
```
if (rootUrl === undefined && entry.fullUrl && entry.fullUrl.length > MAX_BUNDLE_ENTRY_URL_LENGTH) {  // throw error} 
elseif (rootUrl === undefined && enty.fullUrl) {
 rootUrl = entry.fullUrl.match(captureFullUrlParts)
                            ? entry.fullUrl.match(captureFullUrlParts)[1]
                            : undefined;
}
```

src/regExpressions.test.ts
Bingjiling: The actualMatch is actually longer than the expectedMatch, with additional metadata fields like 'input', and those are the fields that we don't care.

nguyen102: Ah, can we use slice then?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice

`expect(actualMatch).toEqual(expectedMatch.slice(expectedMatch.length))`

src/router/bundle/bundleParser.ts
Bingjiling: Seems like CodeQL will still flag it if the two clause are separated completely. I took a middle path of organizing it to two layer of if statement.

src/regExpressions.test.ts
Bingjiling: Emmm that could work. I copied the for loop check from other test and didn't update it.

Bingjiling: Emmm it didn't work actually. The resulted actualMatch result is of type RegExpMatchArray, which is extended from Array. But surprisingly slice does not work on it. I'll keep it as is.

src/regExpressions.test.ts
```diff
@@ -64,3 +70,41 @@ describe('captureFullUrlParts', () => {
...
+    });
+    test('Extra long resource type fail to match', () => {
+        const resourceString = `${'Patient'.repeat(30)}/12345678`;
+        console.log(resourceString);
```
nguyen102: Should this be deleted?

src/regExpressions.test.ts
```diff
@@ -64,3 +70,41 @@ describe('captureFullUrlParts', () => {
...
+        const expectedMatch = ['Patient/12345678', 'Patient'];
+
+        for (let i = 0; i < expectedMatch.length; i += 1) {
+            // @ts-ignore
+            expect(actualMatch[i]).toEqual(expectedMatch[i]);
+        }
+    });
+    test('Extra long resource id fail to match', () => {
+        const resourceString = `Patient/${'12345678'.repeat(14)}`;
+        console.log(resourceString);
```
nguyen102: should this be deleted?

nguyen102 (approved):

